PR #1, opened 2023-01-25
200% default minimum liquidity ratio

noahlitvin:
Seeing `Error: invalid BigNumber string`

When trying to use parseEther helper:
```
[setting.minimum_liquidity_ratio]
defaultValue = "<%= parseEther('200') %>"
```


noahlitvin:
Incorporating this into an incoming PR

PR #22, opened 2023-05-29
enable weth in goerli

fritzschoff:
Screenshot 2023-05-29 at 18 23 18

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ Conduct the following process for each network:
 - Perform a dry-run and confirm that the actions that would be executed by Cannon are expected:
 
 ```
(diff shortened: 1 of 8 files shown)


dbeal-eth (changes requested):
1. Mainnet ci check is failing
2. I would rather not have a setting for a boolean just to identify if weth is enabled or not. We should find a better solution.

dbeal-eth:
It looks like mainnet was failing before this pr. Still, not an excuse for it not working now as the error seems to be directly related to your changes. Let's get the ci all the way working.

fritzschoff:
> 1. Mainnet ci check is failing
> 2. I would rather not have a setting for a boolean just to identify if weth is enabled or not. We should find a better solution.

boolean is fine here. Its pretty clear

cavalier-eth:
@fritzschoff this looks old can it be closed?

PR #43, opened 2023-09-07
Simulate release for every config we have

noisekit:
TEST ALL THNIGS


dbeal-eth (approved):

dbeal-eth:
@noisekit `competition` need to simulate upgrade to the `competition`  preset instead of the default main preset. otherwise it will look like its rebuilding everything when its actually not

regarding failure of `arbitrum-goerli` job, there seems to be issues with the arbitrum network with anvil simulation which are out of our control. we almost alwyas have to YOLO that release out. perhaps we should remove from CI for now?

PR #44, opened 2023-09-08
Indent all things

noisekit:
- Fix indentation (tabs -> spaces) in all tomls
- Normalise to 4 spaces everywhere (looks like that was the standard, but we had 2 and 3 mixed up)
- Add prettier and toml plugin
- Add `Lint` gh action to check prettiness

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,26 @@
+name: Lint
+
+on:
+  pull_request: {}
+  push:
+    branches:
+      - "test"
+
+jobs:
+  lint:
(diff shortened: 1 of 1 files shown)


dbeal-eth (changes requested):
Can't add the zip files to the repo, they are probably too heavy for the cannon deploy app. This repo is supposed to be pretty lean so that it can easily be imported into browser environments.

Sorry

dbeal-eth (approved):

dbeal-eth:
Really good to have prettier on this repo now 👍

PR #47, opened 2023-09-19
Set additional SNX pool configuration

noisekit:
- [x] turn off susd issuance
- [x] not opt into new collateral by default
- [x] not back markets (empty list of markets in pool config)
- [x] unrestricted amount of snx, set collateralLimitD18 - max int
- [x] turn off any other collateral, apart from snx (now andin future)
- [x] remove incorrect snx pool deployment on all networks
- [x] call `removeApprovedPool` for all networks except `goerli`. ID is hardcoded because the original id for the pool was 69. We will probably deploy them at some point in the future but with matching ID.


Rickk137 (approved):

cavalier-eth (approved):
lgtm

jmzwar (approved):
This looks gtm but maybe wait till post testnet trading comp

PR #51, opened 2023-09-29
Upgrade cannon

noisekit:
- update cannon
- reuse repo version of cannon in gh actions instead of globally installing latest cannon


package.json
```diff
@@ -11,9 +11,11 @@
     "pretty:check": "prettier . '**/*.toml' --check"
   },
   "devDependencies": {
-    "@usecannon/cli": "^2.5.4",
+    "@usecannon/cli": "^2.8.1",
```
dbeal-eth, line 14: according to the PR cannon cli should be locked to 2.7.1?

dbeal-eth (approved):

PR #52, opened 2023-10-03
Update configs for ccip

mjlescano:
This PR is necessary for the changes added here: https://github.com/Synthetixio/synthetix-v3/pull/1682


mjlescano:
This is way too old

PR #54, opened 2023-10-04
New deployment with permissioned erc20

noisekit:
Reopening for #50 

@sunnyvempati cannon does not remove things when we remove them from configs. It is append-only. So to disable something we need to add extra commands.

Build must be green before we merge, this ensures our deployments will work.


sunnyvempati:
@noisekit ty.  seems like CI is working well here, i wasn't sure if it was just being flaky.  i'll try to figure out best approach here

noisekit:
@sunnyvempati close?

PR #55, opened 2023-10-09
Unapprove incorrect SNX Passive pool on Goerli

noisekit:
too many pools
_ 2023-10-09 at 16 46 27

Follow up to #47


jmzwar (approved):

cavalier-eth:
no longer needed as config with be redeployed
UI will need to be updated

PR #58, opened 2023-10-13
SIP-324 Passive SNX Pool - Goerli `staging` deployment

noisekit:
1. goerli deployment is a mirror of `mainnet` deployment with goerli-specific configs only
2. after merging/deploying we will transfer ownership to the safe account and deploy in the same way we deploy mainnet.
3. we can only add things to it as a final test before deploying to mainnet.
4. if we want to have `dev` deployment we should use a separate deployment file like `omnibus-goerli-dev.toml`

NOTE: Simulate release for goerli is expected to fail with `Proxy InitialCoreProxy was modified` because of full redeployment from scratch (salt change)

### DIFF
--- a/omnibus-goerli.toml
+++ b/omnibus-goerli.toml
@@ -6,13 +6,16 @@ include = [
     "tomls/core.toml",
     "tomls/permissions.toml",
     "tomls/collaterals/snx.toml",
-    "tomls/collaterals/weth.toml",
     "tomls/pools/spartan-council.toml",
     "tomls/pools/passive-snx.toml",
-    "tomls/permit-all-createPool.toml",
-    "tomls/permit-all-registerMarket.toml",
     "tomls/permit-all-transferCrossChain.toml",
 ]
+
+[setting.target_preset]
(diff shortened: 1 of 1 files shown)


dbeal-eth (changes requested):
Unless the goerli deployment is irreparably busted, we should not be redeploying the main preset deployment to a new salt.

The reason is because our cross chain requirements stipulate that all v3 deployments should have the same address (this is to both reduce complexity of deployment for integeators, reduce gas usage to load other network contract addresses, improve security and reduce configuration complexity by not having to configure other network addresses, etc.)

If you wish to deploy a new preset or salt, please put it in a new omnibus file

dbeal-eth:
If you believe that the gorilla deployment is irreparably busted, please let me know and we can discuss the options. When I was giving you this advice and discord, I thought that you were working on a separate deployment

dbeal-eth (approved):
Was informed all test nets will be being redeployed for fresh state. So with that in mind, changing the salt is correct.

PR #70, opened 2023-11-13
Deps update

noisekit:
- new cannon
- updated simulate-release checks

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,13 +15,13 @@
   "devDependencies": {
     "@ethersproject/bytes": "^5.7.0",
     "@iarna/toml": "^3.0.0",
-    "@usecannon/builder": "^2.9.0",
-    "@usecannon/cli": "^2.9.1",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -15,13 +15,13 @@
   "devDependencies": {
     "@ethersproject/bytes": "^5.7.0",
     "@iarna/toml": "^3.0.0",
-    "@usecannon/builder": "^2.9.0",
-    "@usecannon/cli": "^2.9.1",
+    "@usecannon/builder": "2.9.10",
+    "@usecannon/cli": "2.9.11-alpha.0",
```
bachstatter, line 19: Should we wait until it's not alpha before upgrading?

bachstatter (approved):

package.json
noisekit: Nope we don't it is a fix that required for cannon to work. I have no idea why it is not properly released yet

PR #66, opened 2023-10-31
Base Goerli Andromeda

noisekit:
Replace #65

PR #71, opened 2023-11-13
setup pyth to use imported pyth package

noisekit:
- [x] Add pyth configs but as a separate PR 
- [ ] TODO solve issues with simulate release tests.

Extracted from #66 as it affects other deployments

cc @dbeal-eth


tomls/markets/common/settlement-settings.toml
```diff
@@ -3,7 +3,7 @@
 [setting.pythFeedUrl]
 
 [setting.settlementReward]
-defaultValue = "0"
+defaultValue = "0.1"
```
noisekit, line 6: Because settlement settings are not updatable but append only we cannot just change settings. May I suggest removing this file entirely and inline settings into the addSettlementStrategy invoke instead? Because once we add it we cannot update or remove it. Any change will require adding a new strategy and disabling the old one.

noisekit (commented):
We must *not* change settlement strategy settings.

tomls/markets/perps/btc.toml
```diff
@@ -49,7 +49,7 @@ func = "addSettlementStrategy"
 args = [
     "<%= settings.btcPerpsMarketId %>",
     # strategyType = 0 (pyth)
-    { strategyType = "0", settlementDelay = "<%= settings.bigCapSettlementDelay %>", settlementWindowDuration = "<%= settings.bigCapSettlementWindowDuration %>", priceWindowDuration = "<%= settings.bigCapPriceWindowDuration %>", priceVerificationContract = "<%= settings.pythPriceVerificationAddress %>", feedId = "<%= settings.pythBtcFeedId %>", url = "<%= settings.pythFeedUrl %>", settlementReward = "<%= settings.settlementReward %>", priceDeviationTolerance = "<%= parseEther('1') %>", disabled = false }
+    { strategyType = "0", settlementDelay = "<%= settings.bigCapSettlementDelay %>", settlementWindowDuration = "<%= settings.bigCapSettlementWindowDuration %>", priceWindowDuration = "<%= settings.bigCapPriceWindowDuration %>", priceVerificationContract = "<%= settings.pythPriceVerificationAddress %>", feedId = "<%= settings.pythBtcFeedId %>", url = "<%= settings.pythFeedUrl %>", settlementReward = "<%= parseEther(settings.settlementReward) %>", priceDeviationTolerance = "<%= parseEther('1') %>", disabled = false }
```
noisekit, line 52: I suggest not to wrap settings into `parseEther` to avoid confusion. And instead use `parseEther` in settings itself `defaultValue = "<%= parseEthers() %>"`
This change introduces variability in how settings are used as some of them are already in wei and some are plain numbers.

noisekit: I have somewhat related idea (not yet formed) on how to use something like `zod` schema to validate values, so they do come in at least expected ranges.
This is not part of generic cannon setup but perhaps can be implemented as a synthetix-specific plugin? So we declare, let's say, that settlementReward is expected to be in  wei, and if we detect values that are not in a _reasonable_ wei range - validation fails.
This schema can potentially be half-auto extracted from solidity. And/or at least be collocated with contract itself.
Would appreciate any feedback on the idea so we can at least avoid wildly incorrect valued sent to our contracts by mistake

noisekit:
This is a part of #73

PR #73, opened 2023-11-14
Build perps andromeda

leomassazza:


tomls/markets/perps/btc.toml
```diff
@@ -49,7 +49,7 @@ func = "addSettlementStrategy"
 args = [
     "<%= settings.btcPerpsMarketId %>",
     # strategyType = 0 (pyth)
-    { strategyType = "0", settlementDelay = "<%= settings.bigCapSettlementDelay %>", settlementWindowDuration = "<%= settings.bigCapSettlementWindowDuration %>", priceWindowDuration = "<%= settings.bigCapPriceWindowDuration %>", priceVerificationContract = "<%= settings.pythPriceVerificationAddress %>", feedId = "<%= settings.pythBtcFeedId %>", url = "<%= settings.pythFeedUrl %>", settlementReward = "<%= settings.settlementReward %>", priceDeviationTolerance = "<%= parseEther('1') %>", disabled = false }
+    { strategyType = "0", settlementDelay = "<%= settings.bigCapSettlementDelay %>", settlementWindowDuration = "<%= settings.bigCapSettlementWindowDuration %>", priceWindowDuration = "<%= settings.bigCapPriceWindowDuration %>", priceVerificationContract = "<%= settings.pythPriceVerificationAddress %>", feedId = "<%= settings.pythBtcFeedId %>", url = "<%= settings.pythFeedUrl %>", settlementReward = "<%= parseEther(settings.settlementReward) %>", priceDeviationTolerance = "<%= parseEther('1') %>", disabled = false }
```
noisekit, line 52:
```suggestion
    { strategyType = "0", settlementDelay = "<%= settings.bigCapSettlementDelay %>", settlementWindowDuration = "<%= settings.bigCapSettlementWindowDuration %>", priceWindowDuration = "<%= settings.bigCapPriceWindowDuration %>", priceVerificationContract = "<%= settings.pythPriceVerificationAddress %>", feedId = "<%= settings.pythBtcFeedId %>", url = "<%= settings.pythFeedUrl %>", settlementReward = "<%= settings.settlementReward %>", priceDeviationTolerance = "<%= parseEther('1') %>", disabled = false }
```

This is very confusing to have parseEther here. ParseEthers should be exactly where settings were set so we can see at a glance what actual value it is. wei or not wei.

tomls/markets/common/settlement-settings.toml
```diff
@@ -3,7 +3,7 @@
 [setting.pythFeedUrl]
 
 [setting.settlementReward]
-defaultValue = "0"
+defaultValue = "0.1"
```
noisekit, line 6: Move my comment from #71 

Because settlement settings are not updatable but append only we cannot just change settings. May I suggest removing this file entirely and inline settings into the addSettlementStrategy invoke instead? Because once we add it we cannot update or remove it. Any change will require adding a new strategy and disabling the old one.

tomls/markets/spot/eth.toml
```diff
@@ -38,7 +38,8 @@ func = "updatePriceData"
 args = [
     "<%= extras.synth_eth_market_id %>",
     "<%= extras.eth_oracle_id %>",
-    "<%= extras.eth_oracle_id %>"
+    "<%= extras.eth_oracle_id %>",
```
noisekit, line 41: This change affects other deployments that rely on older version of spot market.

```js
Failure on step invoke.setSynthPriceEth
Error: too many arguments: passed to contract (count=5, expectedCount=3, code=UNEXPECTED_ARGUMENT, version=contracts/5.7.0)
```

noisekit: We need to have this change separated from common configs and specify for andromeda deployment exclusively.

noisekit (approved):
Finally, we're good to go

PR #75, opened 2023-11-16
Setup e2e testing for Andromeda

noisekit:
- [x] make sure we capture chain and preset in deployments
- [x] separate tests for each deployment
- [x] bootstrap e23e test for andromeda (currently just mint fake TKN that's used as `USDC`)

NOTES:
- The first step. It tests against ALREADY PUBLISHED deployment.
- Next step would be to test against the toml file changes (will need to make a fork upgrade deployment with cannon and then test against it)
- Also need to add more steps (essentially repeating what we have in https://github.com/Synthetixio/synthetix-v3/blob/main/sandboxes/andromeda/consumer.cannonfile.toml plus checking limits)


jmzwar (approved):
Ok let's just gtm for the moment and think about enhancements later

PR #77, opened 2023-11-20
Test andromeda max market collateral

noisekit:
Took me some time to figure out that we already have more than $2M of collateral deployed. So soon there will be not that much left to wrap

added error parsing too

_ 2023-11-20 at 15 33 31


jmzwar (approved):
Good to start off

PR #92, opened 2023-12-08
Burn leftover debt

noisekit:
WIP

- Add mining, debt burning, and sUSD swapping
- Make sure test user has some sUSD in the wallet (20)
- Deposit 30 sUSD to the system
- Burn debt ~~(does not seem to work =/)~~ with multicall it works well

image


bachstatter (approved):

jmzwar (approved):

noisekit:
The test will pass after I change ownership for pool owner

PR #99, opened 2023-12-17
support for building andromeda testnet on local

dbeal-eth:
accomplishes this by detecting the chain id of the currently building network. if its local, it replaces problematic oracle nodes with constant nodes

otherwise, the deployed network extremely closely resembles what is actually deployed to testnet, even after upgrades. this would be suitable for building a non-fork sandbox of any network (although right now its only tested with andromeda testnet)


noahlitvin:
Clever. I figured we'd have it use a local build of pyth/op-gas packages and create a new top-level cannonfile, but I like this.

tomls/oracles/pyth-btc.toml
noahlitvin, line 1: Would it be better to swap the staleness circuit breaker node?

leomassazza:
Is it possible to use a mock contract for the we cannot use in non-forked networks instead of changing the nodes architecture? That way the change is minimized (and more accurate to what is expected to happen in real networks) for local tests

PR #103, opened 2023-12-20
Enable perps trading tests for mainnet

noisekit:


e2e/tasks/commitPerpsOrder.js
```diff
@@ -32,7 +32,7 @@ async function commitPerpsOrder({ wallet, accountId, marketId, sizeDelta, settle
     sizeDelta: ethers.utils.parseEther(`${sizeDelta}`),
     settlementStrategyId,
     acceptablePrice: ethers.utils.parseEther(
-      Math.floor(pythPrice * (sizeDelta > 0 ? 2 : 0.5)).toString()
+      Math.floor(pythPrice * (sizeDelta > 0 ? 1.01 : 0.99)).toString()
```
bachstatter, line 35: If price moves 1% or you get a bad fill due to skew this will be a bit flakey. Why do we want this tight slippage protection in tests?

e2e/tests/omnibus-base-goerli-andromeda.toml/Perps_Trading.e2e.js
```diff
@@ -297,19 +297,23 @@ describe(require('path').basename(__filename, '.e2e.js'), function () {
...
     const settlementStrategyId = extras.btc_pyth_settlement_strategy;
     const maxSize = await PerpsMarketProxy.getMaxMarketSize(marketId);
+    log({ marketId, maxSize });
     try {
       await commitPerpsOrder({
         wallet,
         accountId,
         marketId,
-        sizeDelta: parseFloat(formatEther(maxSize)),
+        sizeDelta: parseFloat(ethers.utils.formatEther(maxSize.mul(2))),
```
bachstatter, line 306: Isn't it better to test that the correct max is being enforced?

e2e/tasks/commitPerpsOrder.js
noisekit: used for CLI too, not just tests

e2e/tests/omnibus-base-goerli-andromeda.toml/Perps_Trading.e2e.js
noisekit: because using max does not fail. max + infinitely small value does.

e2e/tasks/commitPerpsOrder.js
bachstatter: Yea ok, with this tight slippage it would be slightly better to fetch the current fill price based on your size and base the acceptable price on that rather than the raw pyth price

e2e/tests/omnibus-base-goerli-andromeda.toml/Perps_Trading.e2e.js
bachstatter: 🤔  How did it pass before?

noisekit: no idea, it was failing on mainnet consistently

e2e/tasks/commitPerpsOrder.js
noisekit: what is the slippage to use? 5% is reasonable?

bachstatter: yea that's probably safer, just want to avoid flakiness

bachstatter (approved):

PR #105, opened 2023-12-28
As per polynomial's request

kaleb-keny:
Executed under
https://goerli.basescan.org/tx/0xbada6a495192e1edd74f30a2b1e9290371c41e4c48acc4694ce90bc7fbdb1690

PR #107, opened 2024-01-08
Based Sepolia

noisekit:
- [x] Copy from goerli
- [x] Update price verification address from pyth
- [x] update fUSDC deployment to have 6 decimals
- [x] include 10x OI increase from #105 which is already deployed to Goerli
- [x] e2e tests added and passing
- [x] CI test with special case of NEW deployment (without upgrade)


cavalier-eth (approved):
approving cos its a testnet release

PR #118, opened 2024-01-19
Docgen: collateral configs

noisekit:
- [x] Collateral configs (duplicates removed)
- [x] Align table styling and formatting across pages
- [x] Move Owners section to the bottom
- [x] Add Link to ABI and Addresses to the top (temporarily until integrated directly into the page)

image

### DIFF
--- a/e2e/docgen/lib/numbers.js
+++ b/e2e/docgen/lib/numbers.js
@@ -1,14 +1,30 @@
 const { ethers } = require('ethers');
 const numbro = require('numbro');
 
-function readableBigWei(bigNumber) {
-  if (bigNumber.eq(ethers.constants.MaxUint256)) {
+function rawValue(value) {
+  if (value instanceof ethers.BigNumber) {
+    if (ethers.constants.MaxUint256.eq(value)) {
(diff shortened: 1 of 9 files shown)


cavalier-eth (approved):
🔥

e2e/docgen/lib/numbers.js
```diff
@@ -0,0 +1,71 @@
...
+}
+
+function readableBigWei(value) {
+  if (!(value instanceof ethers.BigNumber)) {
+    return rawValue(value);
+  }
+  if (ethers.constants.MaxUint256.eq(value)) {
+    return '<code>MaxUint256</code>';
+  }
+  if (ethers.constants.MaxInt256.eq(value)) {
```
bachstatter, line 24: Do be totally correct here I think this is slighlyt better:
`ethers.constants.MaxInt256.eq(value.abs())`
as value could be neg

bachstatter (approved):

e2e/docgen/lib/numbers.js
noisekit: good one, yep will add this fix. Realistically we probably never ever use it. But worth adding anyway

PR #125, opened 2024-01-29
Report gas usage in tests

noisekit:
Currently will generate CSV file for each test run. And keeps that file in CircleCI artifacts.
We can also push it to some repo (maybe docs repo?). But that would be done separately if necessary.

This is a first iteration and only the minimal set of fields is exported:
- `gasUsed`
- `baseFeePerGas`

As we are running this on a fork the values might not be exactly same as they would be on real chain

csv looks like this 

_ 2024-01-29 at 14 18 05

### DIFF
--- a/e2e/debugTarget.js
+++ b/e2e/debugTarget.js
@@ -1,0 +1,14 @@
+require('supports-color'); // for prettier logs
+const debug = require('debug');
+const util = require('util');
+const fs = require('fs');
+
+const stream = process.env.DEBUG_TARGET
+  ? fs.createWriteStream(require('path').resolve(process.env.DEBUG_TARGET), { flags: 'a' })
+  : process.stderr;
+
+function log(...args) {
+  return stream.write(util.format(...args) + '\n');
+}
+
+debug.log = log;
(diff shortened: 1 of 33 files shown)


e2e/debugTarget.js
```diff
@@ -0,0 +1,14 @@
+require('supports-color'); // for prettier logs
+const debug = require('debug');
+const util = require('util');
+const fs = require('fs');
+
+const stream = process.env.DEBUG_TARGET
+  ? fs.createWriteStream(require('path').resolve(process.env.DEBUG_TARGET), { flags: 'a' })
```
noisekit, line 7: If ENV set to the file all the debug logs go there, so we can have some more clean test logs + test failure logs in CI

e2e/gasLog.js
```diff
@@ -0,0 +1,42 @@
+const { ethers } = require('ethers');
+
+const GAS_REPORT = process.env.GAS_REPORT ? require('path').resolve(process.env.GAS_REPORT) : null;
+
+if (GAS_REPORT) {
+  require('fs').writeFileSync(
```
noisekit, line 6: Optional CSV exporting for gas data

davidvuong (approved):

PR #126, opened 2024-01-30
Increase limits 4x of current limit to completely cover any market irregularities

noisekit:
1. need to deposit basically full current market size to avoid market being locked because perps is so much bigger than lp -- 2x
2. debt gets so big so quick I just want to have same amount of snxUSD to burn when undelegaing -- 3x then
3. ~~it works lol~~ how about nope


davidvuong (approved):

cavalier-eth (approved):
value changes lgtm

PR #128, opened 2024-02-12
Add interest rate to andromeda

leomassazza:

PR #135, opened 2024-03-06
increase `maxDebtShareValueD18` for all networks

Tburm:
Increase the `maxDebtShareValueD18` setting, which will increase the credit capacity of the perps markets.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -484,28 +484,12 @@ workflows:
           provider-url: https://sepolia.base.org
 
       - fork-test:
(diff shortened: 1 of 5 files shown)


noisekit:
all goerli are deprecated

noisekit:
Because 3.3.15 was released from #128 branch any fork tests based off main branch will fail now (for base sepolia). When we finalise the release and merge the branch - things will stabilise.
Alternatively you can set base for the PR to `add-interest-rate` branch instead of `main` and wait until it is merged.

PR #150, opened 2024-03-15
Arb mainnet

dbeal-eth:


noisekit (approved):

tomls/collaterals/weth.toml
```diff
@@ -1,4 +1,4 @@
-include = ["../oracles/eth.toml"]
```
noisekit, line 1: Sepolia and polygon failing because of this

noisekit (changes requested):
Cannon change global weth config

dbeal-eth:
I made the change to the global config because we probably should be migrating all the feeds over to pyth anyway

noisekit:
> I made the change to the global config because we probably should be migrating all the feeds over to pyth anyway

Then we need to deprecate and delete old configs for sepolia and polygon or provide a way to migrate. As part of this pr.

tomls/collaterals/weth.toml
noisekit: fixed by removing nested includes

tomls/omnibus-arbitrum-mainnet-arbthetix/perps/feeCollector.toml
```diff
@@ -0,0 +1,71 @@
...
+[setting.pythSnxFeedId]
+defaultValue = "0x39d020f60982ed892abbcd4a06a276a9f9b7bfbce003204c110b6e488f502da3"
+
+[provision.buyback_snx]
+source = "<%= settings.buyback_snx_package %>"
+targetPreset = "<%= settings.target_preset %>"
+options.premium = "<%= settings.buyback_premium %>"
+options.snx_fee_share = "<%= settings.buyback_fee_share %>"
+options.oracle_manager_address = "<%= imports.system.imports.oracle_manager.contracts.Proxy.address %>"
+options.snx_token = "<%= imports.snx_mock_collateral.contracts.MintableToken.address %>"
```
noisekit, line 21: we have plenty of references to `mock` in these configs.

noisekit (approved):
Fixed outstanding issues and make this PR non-breaking. Plus fixed few more copy-paste issues.

As it is deployed now and there is no cannon diff anymore - we can merge and all other changes will need new PRs.

PR #153, opened 2024-03-18
Idempotent Setters Only

noahlitvin:
AFAIK, all markets only have 1 settlement strategy, which means its been assigned to ID `0`.

We should be able to remove the "add" steps and hardcode this ID, such that updated settings files don't trigger new settlement strategies to be added.

Strategies can be added in the future by adding a new "set" step with ID 1, 2 ,3, etc.


noisekit (approved):
Looks good as long as we get CI happy

noahlitvin:
Bumping OP version resolved that CI test because it adds the `setSettlementStrategy` function.

noisekit (approved):

PR #166, opened 2024-03-25
Remove sepolia

noahlitvin:


omnibus-sepolia.toml
```diff
@@ -1,64 +0,0 @@
-name = "synthetix-omnibus"
```
noisekit, line 1: you cannot just delete a deployment and not comment out CI tests for it.
Please wait for CI to get green before merging. Because now all branches and main branch are red.

PR #167, opened 2024-03-25
Add ETH Sepolia

noahlitvin:
* Waiting on Pool/Router addresses from Chainlink before merging
* This should be run without `upgrade-from` (Should be a fresh deployment, do we need to change salt?)


omnibus-sepolia.toml
```diff
@@ -0,0 +1,64 @@
...
+
+[setting.ccip_token_pool]
+defaultValue = "0x9b65749b38278060c5787cce0391ac7f1094c8e8"
+
+# TODO clean this block after deployment
+[invoke.unapprove_wrongfully_deployed_snx_pool]
+target = ["system.CoreProxy"]
+fromCall.func = "owner"
+func = "removeApprovedPool"
+args = ["2"]
```
noisekit, line 64: this is very unnecessary.
```suggestion
```

omnibus-sepolia.toml
```diff
@@ -0,0 +1,64 @@
...
+    "tomls/oracles/eth.toml",
+    "tomls/collaterals/weth.toml",
+    "tomls/pools/spartan-council.toml",
+    "tomls/permit-all-createPool.toml",
+    "tomls/permit-all-registerMarket.toml",
+    "tomls/permit-all-transferCrossChain.toml",
+]
+
+[setting.snx_package]
+defaultValue = "synthetix:3.3.5"
```
noisekit, line 18: shouldn't we update to the latest?

dbeal-eth: no it should match mainnet I think to start?

noisekit: Yeah. This is just a revert of deleted deployment. Did not realise that at the time

omnibus-sepolia.toml
```diff
@@ -0,0 +1,57 @@
+name = "synthetix-omnibus"
```
noisekit, line 1: Actually... lets enable test too before merge please. This should be an upgrade

PR #170, opened 2024-03-27
Base mainnet perps v3.1

sunnyvempati:


tomls/omnibus-base-mainnet-andromeda/perps/btc-invokes.toml
```diff
@@ -0,0 +1,139 @@
+include = ["../../oracles/btc.toml", "../../markets/common/bigcap-settings.toml"]
```
noisekit, line 1: no nested includes. Include in root file

sunnyvempati: can do this in a separate PR for now; was just following what base sepolia had

noisekit: Yes please. Current state of multilevel multidirectional includes is pretty poor and it is very hard to find out what is going on. And easy to make a mistake.

For any new code please keep includes only at the very top level toml.

PR #174, opened 2024-04-03
Cannon v2.12

noahlitvin:


tomls/omnibus-base-mainnet-andromeda/perps/btc-invokes.toml
```diff
@@ -67,6 +67,7 @@ args = [
...
     { strategyType = "0", settlementDelay = "<%= settings.bigCapSettlementDelay %>", settlementWindowDuration = "<%= settings.bigCapSettlementWindowDuration %>", priceVerificationContract = "<%= imports.pyth_erc7412_wrapper.contracts.PythERC7412Wrapper.address %>", feedId = "<%= settings.pythBtcFeedId %>", settlementReward = "<%= parseEther(settings.settlementReward) %>", disabled = false, commitmentPriceDelay = "<%= settings.commitmentPriceDelay %>" },
 ]
+depends = ['provision.perpsFactory', 'setting.btcPerpsMarketId', 'invoke.addPerpsBtcSettlementStrategy', 'provision.pyth_erc7412_wrapper', 'setting.bigCapSettlementDelay', 'setting.bigCapSettlementWindowDuration', 'setting.pythBtcFeedId', 'setting.settlementReward']
```
noisekit, line 70: This mean we have a breaking change in new version of cannon. Can we make sure it does not?

noahlitvin: We found an edge case where the inferred dependencies would silently fail, so we've required this to be made explicit out of an abundance of caution.

dbeal-eth (approved):
Conditional on CI pass

PR #177, opened 2024-04-05
bump version

sunnyvempati:


leomassazza (approved):

omnibus-base-mainnet-andromeda.toml
```diff
@@ -1,5 +1,5 @@
 name = "synthetix-omnibus"
-version = "9"
+version = "13"
```
noisekit, line 2: why not 10?

sunnyvempati: there's a `10` on sepolia already, so didn't want it to be confusing

PR #182, opened 2024-04-07
SCCP-312: Increase Base ETH and BTC OI limits to $20m

noisekit:
- [SCCP-312: Increase Base ETH and BTC OI limits to $20m](https://sips.synthetix.io/sccp/sccp-312/)
- Updated tests
- Bumped version

UPD: as we want to maintain LP/OI 1/2 proportion we should get #183 or #184 + wait first

PR #183, opened 2024-04-07
Half of SCCP-312: Increase Base ETH and BTC OI limits to $20m

noisekit:
Same as #182 but half as a step one as we want to maintain 1/2 proportion of LP/OI


cavalier-eth:
lets go with the full $20m instead from https://github.com/Synthetixio/synthetix-deployments/pull/182

cavalier-eth:
closing

PR #189, opened 2024-04-10
SIP-373 and SIP-376 Sepolia tomls

leomassazza:


noisekit (approved):
lgtm

tomls/markets/perps/arb.toml
```diff
@@ -0,0 +1,119 @@
+include = [
+    # TODO: this is a funny necessary dependency because the oracle manager is registering the same node twice
+    # however, it will not emit "NodeRegistered" the second time because its an unnecessary call
+    "../../oracles/pyth-arb.toml",
+    "../common/bigcap-settings.toml",
```
noisekit, line 5: these are not big caps, aren't them

noisekit (commented):
Need e2e test for each new market too.

noisekit:
pumping so many txns in tests now that blockchain is more than 2min ahead of real time.
tests fail because we have 2min timeout in mocha.
All these txns are from setting up all the new markets on fork. And each txn mines the block.

image

PR #195, opened 2024-04-15
Fix tests after SCCP-320

leomassazza:
After #193 and #192 we have some failing tests.


noisekit (approved):

noisekit:
for future: cannon needs to be in a separate PR

PR #199, opened 2024-04-22
make numbers pretty - mainnet

kaleb-keny:


leomassazza (approved):

omnibus-base-mainnet-andromeda.toml
```diff
@@ -1,5 +1,5 @@
 name = "synthetix-omnibus"
-version = "20"
+version = "21"
```
noisekit, line 2: Same here, no need to bump version. For cannon this update is noop
```suggestion
version = "20"
```

noisekit (approved):
without version bump all good.

PR #200, opened 2024-04-22
Make big numbers readable sepolia

kaleb-keny:


kaleb-keny:
hey @noisekit i will push this, verify that no parameters were changed by comparing to mainnet then push mainnet, unfortunately the script says it's updating stuff, but nothing is being updated.
" Hmm yea, I think we'd want it to hash the steps after the template is compiled to avoid this, but I'd be concerned there would be unintended side-effects in that case." As per noah

omnibus-base-sepolia-andromeda.toml
```diff
@@ -1,5 +1,5 @@
 name = "synthetix-omnibus"
-version = "20"
+version = "21"
```
noisekit, line 2: I don't believe we need to up this version as this change is a noop, purely cosmetics.
```suggestion
version = "20"
```

noisekit (approved):
minus the version bump - much better now, those bonk numbers omg

tomls/omnibus-base-sepolia-andromeda/perps/pepe.toml
```diff
@@ -5,7 +5,7 @@ defaultValue = "0xd69731a2e74ac1ce884fc3890f7ee324b6deb66147055249568869ed700882
 defaultValue = "1200"
 
 [setting.perpsPepeSkewScale]
-defaultValue = "8400000000000"
+defaultValue = "<%= String(8_400_000_000_000) %>"
```
noisekit, line 8: bonkers

kaleb-keny: thank you for catching that one

PR #203, opened 2024-04-23
Sepolia new markets (SIPs 370 to 380)

leomassazza:


e2e/tasks/doAllPriceUpdatesSepolia.js
```diff
@@ -0,0 +1,201 @@
+#!/usr/bin/env node
```
noisekit, line 1: This is unnecessary because `doPriceUpdate` for market that does not exist is a noop and ok to do. So global `doAllPriceUpdates` should work fine for any market (that exists or does not exist)

noisekit (changes requested):
remove `doAllPriceUpdatesSepolia` and update global `doAllPriceUpdates`

noisekit (approved):
Thanks, looking good

kaleb-keny:
God i love v3 & cannon, merging and sending, thank you guys

PR #208, opened 2024-04-29
Deployments Dashboard

noisekit:


e2e/dashboard.js
```diff
@@ -0,0 +1,134 @@
...
+  return out.join('\n');
+}
+
+module.exports = {
+  dashboard,
+};
+
+if (require.main === module) {
+  require('./inspect');
+  dashboard().then(prettyMd).then(console.log);
```
0xjocke, line 133: Dashboard return html string but we call prettyMd?

.circleci/config.yml
```diff
@@ -460,6 +460,62 @@ jobs:
...
+      - store_artifacts:
+          path: "./e2e/deployments"
+
+      - run:
+          name: "Generate docs"
+          environment:
+            DEBUG: "e2e:*"
+          command: |
+            mkdir -p docs/for-developers
+            node e2e/dashboard > docs/for-developers/deployment-info.md
```
0xjocke, line 490: Same thing here, shouldn't this be `.html`

e2e/dashboard.js
noisekit: No it does not. It returns MD that has html in it for correct formatting.

.circleci/config.yml
noisekit: it is md not html.

0xjocke (approved):

PR #210, opened 2024-05-06
SIP-367: Synthetix V3 Deployment to Arbitrum (Sepolia testnet)

noisekit:
Cannon build succeeds
image


omnibus-arbitrum-sepolia.toml
```diff
@@ -0,0 +1,186 @@
...
+target = ["system.CoreProxy"]
+fromCall.func = "getPoolOwner"
+fromCall.args = ["<%= settings.sc_pool_id %>"]
+func = "setPoolConfiguration"
+args = [
+    "<%= settings.sc_pool_id %>",
+    [
+        { marketId = "<%= extras.synth_dai_market_id %>", weightD18 = 1, maxDebtShareValueD18 = "<%= parseEther('1') %>" },
+        { marketId = "<%= extras.synth_usdc_market_id %>", weightD18 = 1, maxDebtShareValueD18 = "<%= parseEther('1') %>" },
+        { marketId = "<%= imports.perpsFactory.settings.superMarketId %>", weightD18 = 1, maxDebtShareValueD18 = "<%= parseEther('1') %>" },
```
noahlitvin, line 172: Probably want significantly more weight for perps market over the spot markets/wrappers

omnibus-arbitrum-sepolia.toml
```diff
@@ -0,0 +1,186 @@
...
+defaultValue = "<%= parseEther('0.000001') %>"
+
+[setting.settlementMaxRoundingLoss]
+defaultValue = "<%= parseEther('0.000001') %>"
+
+[setting.priceDeviationTolerance]
+defaultValue = "<%= parseEther('1') %>"
+
+[setting.spotStrictStalenessTolerance]
+defaultValue = "60"
```
noahlitvin, line 96: This can probably be looser (or bigCapDefaultStalenessTolerance, whatever value is being used by the spot market to determine the value)

omnibus-arbitrum-sepolia.toml
```diff
@@ -0,0 +1,186 @@
...
+description = "https://pyth.network/developers/price-feed-ids#pyth-evm-mainnet"
+
+[setting.synthUsdcMaxMarketCollateral]
+defaultValue = "<%= parseEther(String(500_000)) %>"
+
+[setting.synthUsdcAsyncFixedFee]
+defaultValue = "<%= parseEther('0.1') %>" # FIXME
+
+[setting.synthUsdcSkewScale]
+defaultValue = "<%= parseEther(String(100_000)) %>" # FIXME
```
noahlitvin, line 113: Skew scale/fee shouldn't be necessary here

omnibus-arbitrum-sepolia.toml
```diff
@@ -0,0 +1,186 @@
...
+
+[setting.pythUsdcFeedId]
+defaultValue = "0xeaa020c61cc479712813461ce153894a96a6c00b21ed0cfc2798d1f9a9e9c94a"
+description = "https://pyth.network/developers/price-feed-ids#pyth-evm-mainnet"
+
+[setting.synthUsdcMaxMarketCollateral]
+defaultValue = "<%= parseEther(String(500_000)) %>"
+
+[setting.synthUsdcAsyncFixedFee]
+defaultValue = "<%= parseEther('0.1') %>" # FIXME
```
noahlitvin, line 110: We probably don't want a fee here. Also need to verify whether we need max_uint fee on atomic swaps to turn them off

tomls/omnibus-arbitrum-sepolia/spot/dai.toml
```diff
@@ -0,0 +1,80 @@
...
+]
+
+[invoke.addSynthSettlementStrategyDai]
+target = ["spotFactory.SpotMarketProxy"]
+fromCall.func = "getMarketOwner"
+fromCall.args = ["<%= extras.synth_dai_market_id %>"]
+func = "addSettlementStrategy"
+args = [
+    "<%= extras.synth_dai_market_id %>",
+    { strategyType = "1", settlementDelay = "0", settlementWindowDuration = "1", priceVerificationContract = "0x0000000000000000000000000000000000000000", feedId = "0x0000000000000000000000000000000000000000000000000000000000000000", url = "", settlementReward = 0, minimumUsdExchangeAmount = "0", maxRoundingLoss = "0", priceDeviationTolerance = "0", disabled = false },
```
noahlitvin, line 52: Should at least verify maxRoundingLoss = 0 is good here.

omnibus-arbitrum-sepolia.toml
noisekit: What value "disables" skew scale exactly?

omnibus-arbitrum-sepolia.toml
noisekit: So?
```suggestion
defaultValue = "0"
```

omnibus-arbitrum-sepolia.toml
noisekit: What value is needed? 10? 100?

tomls/omnibus-arbitrum-sepolia/spot/dai.toml
noisekit: Deploy is successful. Then it is set to non zero with next invoke

omnibus-arbitrum-sepolia.toml
noisekit: The setting is strict and big cap strict is 60. 
this is what was set for snx spot on OP. 
why would we switch to default (3600) here?

tomls/omnibus-arbitrum-sepolia/spot/dai.toml
sunnyvempati: yep makes sense.  should be a very small value, def not 0 which i think is being set in the next invoke 👍🏽

omnibus-arbitrum-sepolia.toml
sunnyvempati: i dont think we need to invoke the call to set this fee at all tbh.  shouldn't be any fees on both atomic and async for usdc since the price is just 1 right? please confirm @noahlitvin

omnibus-arbitrum-sepolia.toml
noahlitvin: DAI/USDC is expected to fluctuate significantly less than other assets, so just making sure we have a price that's as recent as like a day (maybe even a week?) should be okay? Like if there's a big depeg, presumably someone will be incentivized to post a more recent price to Pyth. It's a risk parameter though, not completely sure the best way to pick a setting, just intuitively seems like it should be looser than anything we've used so far.

omnibus-arbitrum-sepolia.toml
noahlitvin: Just not setting it should be fine, effectively leaving the scale at zero. See https://github.com/Synthetixio/synthetix-v3/tree/main/markets/spot-market#auto-rebalancing-skew-fee

omnibus-arbitrum-sepolia.toml
noahlitvin: The plan is to use a market price for USDC/DAI for this deployment. (Base "settles" in USDC, but this one issues a stablecoin backed by a mix of LP collateral.)

omnibus-arbitrum-sepolia.toml
noahlitvin: I'm not entirely sure. The question is basically what proportion of the collateral should be dedicated to backing sUSDC and sDAI that doesn't have corresponding wrapped assets. Presumably little of this would exist. Maybe kick up line 172 to `98` for now, so its 1% for DAI and 1% for USDC, and we can have some others weigh in.

noisekit: There is a code suggestion in github comment. Very useful
image

```suggestion
        { marketId = "<%= imports.perpsFactory.settings.superMarketId %>", weightD18 = 98, maxDebtShareValueD18 = "<%= parseEther('1') %>" },
```

omnibus-arbitrum-sepolia.toml
noisekit: Makes sense. Intuitively. But *practically* what this value should be?

tomls/omnibus-arbitrum-sepolia/spot/dai.toml
```diff
@@ -0,0 +1,80 @@
...
+    "<%= extras.synth_dai_settlement_strategy_id %>",
+    { strategyType = "1", settlementDelay = "<%= settings.bigCapSettlementDelay %>", settlementWindowDuration = "<%= settings.bigCapSettlementWindowDuration %>", priceVerificationContract = "<%= settings.pyth_price_verification_address %>", feedId = "<%= settings.pythDaiFeedId %>", url = "<%= settings.pythFeedUrl %>", settlementReward = 0, minimumUsdExchangeAmount = "<%= settings.settlementMinimumUsdExchangeAmount %>", maxRoundingLoss = "<%= settings.settlementMaxRoundingLoss %>", priceDeviationTolerance = "<%= settings.priceDeviationTolerance %>", disabled = false },
+]
+
+[invoke.setSynthAsyncFixedFeeDai]
+target = ["spotFactory.SpotMarketProxy"]
+fromCall.func = "getMarketOwner"
+fromCall.args = ["<%= extras.synth_dai_market_id %>"]
+func = "setAsyncFixedFee"
+args = ["<%= extras.synth_dai_market_id %>", "<%= settings.synthDaiAsyncFixedFee %>"]
```
noisekit, line 73: NOTE that we do **NOT** need to set `depends` for these steps because there is no more `parseEther`. It is moved to much better suited place: 
```toml
[setting.synthDaiAsyncFixedFee]
defaultValue = "<%= parseEther('0.1') %>"
```

omnibus-arbitrum-sepolia.toml
noisekit: settled to use separate tolerance settings and both set to 3600

omnibus-arbitrum-sepolia.toml
noisekit: removed skew configuration step

omnibus-arbitrum-sepolia.toml
noisekit: remove fee config step

tomls/omnibus-arbitrum-sepolia/spot/dai.toml
noisekit: I've set it to `1` minimal non-zero value

cavalier-eth (approved):

cavalier-eth:
approved given its only testnet

PR #212, opened 2024-05-06
implements sccp 328 - mainnet

kaleb-keny:


noisekit (approved):
matches the sip 👌

e2e/tests/omnibus-base-mainnet-andromeda.toml/Perps_BTC.e2e.js
```diff
@@ -128,7 +128,7 @@ describe(require('path').basename(__filename, '.e2e.js'), function () {
 
     assert.equal(
       Number(ethers.utils.formatEther(params.initialMarginRatioD18)),
-      13.35,
+      1.54,
```
noisekit, line 131: tbh I would just nuke these checks from test file. Maybe leave only crucial things. Or just nuke these tests entirely. Does not seem awfully useful

kaleb-keny: yeah i think double input is the usefulness. to really ensure we dont have a copy pasta error. But would prefer "smarter" tests maybe. I think we can work on those some day. We dont really need market data to know if the numbers are reasonable or not.

PR #213, opened 2024-05-06
implements sccp 328 - sepolia

kaleb-keny:


omnibus-base-sepolia-andromeda.toml
```diff
@@ -1,5 +1,5 @@
 name = "synthetix-omnibus"
-version = "26"
+version = "25"
```
leomassazza, line 2: Is this downgrade ok?

kaleb-keny: hey yeah it's fine, i think we can stick to 25, i merged a pr that bumped from 24 to 25 earlier https://github.com/Synthetixio/synthetix-deployments/pull/214/files#diff-79a0e3d1b709a381e9fa2ff52898a3914331311a8abbe2b459fb4feed690ed4cL2

kaleb-keny: since i'm planning to deploy both simultaneously...

leomassazza (approved):

PR #219, opened 2024-05-09
Add `target` to mock tokens

noahlitvin:
We're currently publishing to the `mintable-token` namespace. I think we should register `synthetix-mock-tokens` and then merge this change. Thoughts?

It's retriggering the config collateral (probably because it sees the hash of the clone step change) but I think it's fine so long as the address isn't changing? https://app.circleci.com/pipelines/github/Synthetixio/synthetix-deployments/7312/workflows/c1939abb-8104-403f-b11e-3afcd96f35ed/jobs/48043/parallel-runs/0/steps/0-111


omnibus-arbitrum-sepolia.toml
```diff
@@ -163,13 +163,15 @@ description = "https://pyth.network/developers/price-feed-ids#pyth-evm-mainnet"
 
 [provision.dai_mock_collateral]
 source = "mintable-token:1.8"
+target = "synthetix-mock-tokens:1.8@dai"
```
noisekit, line 166: Is this something new? Like a new cannon feature we spoke about before? Would be nice to have some context/links in the PR description.

noisekit: The main question is here - what effect are we expecting? What will happen when we publish? I can see from the build that there are no new contracts deployed, so my guess is that we would publish multple new packages on cannon publish?

noahlitvin: Sorry, just saw this. The idea here is that if/when this omnibus package is published, the subpackage would get published as `synthetix-mock-tokens:1.8@dai` instead of `mintable-token:1.8`. Renaming this operation to `clone` from `provision`, the idea is we "clone" the mintable-token package into a new synthetix-mock-tokens package.

noisekit (approved):

PR #220, opened 2024-05-10
Add tests for borrowing and withdrawing USDh

noisekit:


e2e/tests/omnibus-arbitrum-sepolia.toml/Liquidity_Provider_ARB.e2e.js
```diff
@@ -44,6 +48,11 @@ describe(require('path').basename(__filename, '.e2e.js'), function () {
     await syncTime();
   });
 
+  it('should disable withdrawal timeout', async () => {
+    await setConfigUint({ key: 'accountTimeoutWithdraw', value: 0 });
+    assert.equal(await getConfigUint('accountTimeoutWithdraw'), 0);
+  });
```
Tburm, line 54: I think you're disabling the timeout so that the USDh withdrawal will work. Just noting that this doesn't test any scenario where there is an issue with the timeout, but does test if withdrawals work. Is there a way to do a fast-forward instead?

Tburm (approved):

e2e/tests/omnibus-arbitrum-sepolia.toml/Liquidity_Provider_ARB.e2e.js
Tburm: Not a blocking question, I approved but wondered if you have considered this.

noisekit: We are going to remove timeouts anyway as far as I know so this should be ok

PR #222, opened 2024-05-10
SIP-367: Arbitrum Sepolia - update price verification contract

Tburm:
Update the price verification contract from the Pyth address to the wrapper. With the current settings async orders will fail, since you do not get an `OracleDataRequired` error during order settlement. Instead it simply reverts since the price data doesn't exist.


noisekit (approved):
This is great find, thank you 🙏

noisekit:
Need to bump version too

PR #227, opened 2024-05-14
bump synthetix on andromeda for rewardmanager fix

dbeal-eth:


noisekit (approved):
Looks good to me.

omnibus-base-mainnet-andromeda.toml
```diff
@@ -1,5 +1,5 @@
 name = "synthetix-omnibus"
-version = "26"
+version = "27"
```
noisekit, line 2: Just checking if we have done testnet already?

noisekit: If not, then lets do base sepolia first

dbeal-eth: That's right we haven't

noisekit (approved):

PR #230, opened 2024-05-20
Bfp market carina

0xjocke:
This is the initial work to get l1perp on a testnet. 
- We might have to split things out more so it can be reused by the mainnet omnibus
- And we might not use a preset at all.
 
 But this should be good for an initial testnet release

### DIFF
--- a/omnibus-sepolia-carina.toml
+++ b/omnibus-sepolia-carina.toml
@@ -156,4 +156,26 @@ depends = [
 
 # --- Configure synths markets --- #
 
+# swsteth spot synth
+
+# seth spot synth
+
 # --- Configure bfp markets --- #
+
+# ethperp
+
+# --- Configure SC pool after market creation--- #
+
+# [invoke.set_pool_configuration_sc]
+# target = ["system.CoreProxy"]
+# fromCall.func = "getPoolOwner"
+# fromCall.args = ["<%= settings.sc_pool_id %>"]
+# func = "setPoolConfiguration"
+# args = [
+#     "<%= settings.sc_pool_id %>",
+#     [
+#         { marketId = "<%= extras.synth_eth_market_id %>", weightD18 = 1, maxDebtShareValueD18 = "<%= parseEther('1') %>" },
+#         { marketId = "<%= extras.synth_wsteth_market_id %>", weightD18 = 1, maxDebtShareValueD18 = "<%= parseEther('1') %>" },
+#         { marketId = "<%= extras.eth_perp_market_id %>", weightD18 = 1, maxDebtShareValueD18 = "<%= parseEther('1') %>" },
+#     ],
+# ]
--- a/omnibus-sepolia-carina.toml
+++ b/omnibus-sepolia-carina.toml
@@ -76,7 +76,7 @@ defaultValue = "0x694AA1769357215DE4FAC081bf1f309aDC325306"
 description = "https://docs.chain.link/data-feeds/price-feeds/addresses/?network=ethereum&page=1"
 
 [setting.weth_liquidation_reward]
-defaultValue = "<%= parseEther('10') %>"
+defaultValue = "<%= parseEther('0.005') %>"
 
 [setting.minimum_liquidity_ratio]
 defaultValue = "<%= parseEther('2') %>"
@@ -99,8 +99,14 @@ defaultValue = "<%= parseEther('1.5') %>"
 [setting.wsteth_liquidation_reward_d18]
 defaultValue = "<%= parseEther('0.01') %>"
 
-[setting.wstethMinDelegationD18]
+[setting.wsteth_min_delegation_d18]
 defaultValue = "<%= parseEther('0.01') %>"
+
+[setting.weth_max_pool_collateral]
+defaultValue = "<%= parseEther(String(20_000)) %>"
+
+[setting.wsteth_max_pool_collateral]
+defaultValue = "<%= parseEther(String(20_000)) %>"
 
 # --- BFP configurations  -- #
 #
@@ -111,10 +117,10 @@ defaultValue = "<%= parseEther('0.01') %>"
 defaultValue = "<%= parseEther(String(100_000_000)) %>"
 
 [setting.weth_max_collateral]
-defaultValue = "<%= parseEther(String(1_000_000)) %>"
+defaultValue = "<%= parseEther(String(10_000)) %>"
 
 [setting.wsteth_max_collateral]
-defaultValue = "<%= parseEther(String(1_000_000)) %>"
+defaultValue = "<%= parseEther(String(10_000)) %>"
 
 [setting.susd_skew_scale]
 defaultValue = "0" # skew scale is not used for sUSD.
@@ -227,7 +233,9 @@ depends = ["provision.system"]
 [provision.wsteth_to_steth_ratio_oracle]
 source = "<%= settings.wsteth_to_steth_ratio_oracle %>"
 targetPreset = "<%= settings.target_preset %>"
+options.owner = "<%= settings.owner %>"
 options.wsteth = "<%= settings.wsteth_address %>"
+options.salt = "<%= settings.salt %>"
 
 [invoke.register_node_wsteth_to_steth_ratio_oracle]
 target = ["system.oracle_manager.Proxy"]
@@ -469,7 +477,7 @@ oracleNodeId = "<%= settings.wsteth_oracle_node_id %>"
 issuanceRatioD18 = "<%= settings.wsteth_issuance_ratio_d18 %>"
 liquidationRatioD18 = "<%= settings.wsteth_liquidation_ratio_d18 %>"
 liquidationRewardD18 = "<%= settings.wsteth_liquidation_reward_d18 %>"
-minDelegationD18 = "<%= settings.wstethMinDelegationD18 %>"
+minDelegationD18 = "<%= settings.wsteth_min_delegation_d18 %>"
 depositingEnabled = false
 
 # --- Configure maximum market collateral in Core --- #
@@ -483,7 +491,7 @@ func = "configureMaximumMarketCollateral"
 args = [
     "<%= settings.eth_market_id %>",
     "<%= settings.weth_address %>",
-    "<%= settings.weth_max_collateral %>",
+    "<%= settings.weth_max_pool_collateral %>",
 ]
 
 [invoke.wsteth_maximum_market_collateral]
@@ -493,7 +501,7 @@ func = "configureMaximumMarketCollateral"
 args = [
     "<%= settings.eth_market_id %>",
     "<%= settings.wsteth_address %>",
-    "<%= settings.wsteth_max_collateral %>",
+    "<%= settings.wsteth_max_pool_collateral %>",
 ]
 
 # --- Register Reward Distributors with Core --- #
--- a/omnibus-sepolia-carina.toml
+++ b/omnibus-sepolia-carina.toml
@@ -90,10 +90,23 @@ defaultValue = "1"
 [setting.pool_owner]
 defaultValue = "0x48914229deDd5A9922f44441ffCCfC2Cb7856Ee9"
 
+[setting.wstethIssuanceRatioD18]
+defaultValue = "<%= parseEther('3') %>"
+
+[setting.wstethLiquidationRatioD18]
+defaultValue = "<%= parseEther('1.5') %>"
+
+[setting.wstethLiquidationRewardD18]
+defaultValue = "<%= parseEther('0.01') %>"
+
+[setting.wstethLMinDelegationD18]
+defaultValue = "<%= parseEther('0.01') %>"
+
 # --- BFP configurations  -- #
 #
 #
 
+# Collaterals
 [setting.susd_max_collateral]
 defaultValue = "<%= parseEther(String(100_000_000)) %>"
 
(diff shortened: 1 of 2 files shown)


omnibus-sepolia-carina.toml
```diff
@@ -0,0 +1,436 @@
...
+keeperCancellationGasUnits = 600000
+keeperFlagGasUnits = 1200000
+keeperLiquidateMarginGasUnits = 1200000
+keeperLiquidationGasUnits = 1200000
+keeperLiquidationFeeUsd = "<%= parseEther('5') %>"
+keeperLiquidationEndorsed = "0x48914229deDd5A9922f44441ffCCfC2Cb7856Ee9" # TODO, this is currently owner, change to a keeper
+collateralDiscountScalar = "<%= parseEther('1') %>"
+minCollateralDiscount = "<%= parseEther('0.01') %>"
+maxCollateralDiscount = "<%= parseEther('0.05') %>"
+sellExactInMaxSlippagePercent = "<%= parseEther('0.05') %>"
```
davidvuong, line 261: Can be removed after https://github.com/Synthetixio/synthetix-v3/pull/2146/files is merged and released

omnibus-sepolia-carina.toml
```diff
@@ -0,0 +1,436 @@
...
+defaultValue = "<%= parseEther('100000') %>"
+
+[setting.wsteth_max_collateral]
+defaultValue = "<%= parseEther('100000') %>"
+
+[setting.weth_skew_scale]
+defaultValue = "<%= parseEther('1000000') %>"
+
+[setting.wsteth_skew_scale]
+defaultValue = "<%= parseEther('1000000') %>"
```
davidvuong, line 106: Minor: `1_000_000` for readability

omnibus-sepolia-carina.toml
```diff
@@ -0,0 +1,436 @@
...
+maxInitialMarginRatio = "<%= parseEther('0.9') %>"
+liquidationRewardPercent = "<%= parseEther('0.00002') %>"
+liquidationLimitScalar = "<%= parseEther('1') %>"
+liquidationWindowDuration = 30
+liquidationMaxPd = "<%= parseEther('0.0016') %>"
+
+# --- Configure collateral for BFP --- #
+#
+#
+# TODO figure out how to get sUSD address
```
davidvuong, line 321: Something like `system.USDProxy.address`? System belongs in core.toml, unsure of the exact cannon reference syntax/path.

omnibus-sepolia-carina.toml
```diff
@@ -0,0 +1,436 @@
...
+minKeeperFeeUsd = "<%= parseEther('15') %>"
+maxKeeperFeeUsd = "<%= parseEther('500') %>"
+keeperProfitMarginUsd = "<%= parseEther('20') %>"
+keeperProfitMarginPercent = "<%= parseEther('0.2') %>"
+keeperSettlementGasUnits = 1200000
+keeperCancellationGasUnits = 600000
+keeperFlagGasUnits = 1200000
+keeperLiquidateMarginGasUnits = 1200000
+keeperLiquidationGasUnits = 1200000
+keeperLiquidationFeeUsd = "<%= parseEther('5') %>"
```
davidvuong, line 256: This can be removed too.

omnibus-sepolia-carina.toml
```diff
@@ -0,0 +1,436 @@
...
+    "<%= settings.sc_pool_id %>",
+    "<%= settings.weth_address %>",
+    "<%= extras.weth_rewards_distributor %>",
+]
+
+# --- Configure SC pool after market creation--- #
+#
+#
+
+[invoke.set_pool_configuration_sc]
```
davidvuong, line 426: Much easier without spot market 👍

davidvuong (commented):
Left a few comments

omnibus-sepolia-carina.toml
0xjocke: Yep figures it out!

omnibus-sepolia-carina.toml
0xjocke: Yea we still might want some spot markets here though

davidvuong (approved):

omnibus-sepolia-carina.toml
```diff
@@ -0,0 +1,539 @@
...
+# --- Synthetix pool settings -- #
+#
+#
+
+[setting.eth_aggregator_address]
+defaultValue = "0x694AA1769357215DE4FAC081bf1f309aDC325306"
+description = "https://docs.chain.link/data-feeds/price-feeds/addresses/?network=ethereum&page=1"
+
+[setting.weth_liquidation_reward]
+defaultValue = "<%= parseEther('10') %>"
```
noisekit, line 79: 10 weth reward is a lot

noisekit, line 79:
```suggestion
defaultValue = "<%= parseEther('0.005') %>"
```

omnibus-sepolia-carina.toml
```diff
@@ -0,0 +1,539 @@
...
+# --- BFP configurations  -- #
+#
+#
+
+# Collaterals
+[setting.susd_max_collateral]
+defaultValue = "<%= parseEther(String(100_000_000)) %>"
+
+[setting.weth_max_collateral]
+defaultValue = "<%= parseEther(String(1_000_000)) %>"
```
noisekit, line 114: 1mil of ETH is a lot, what is it now like $4b?

omnibus-sepolia-carina.toml
```diff
@@ -0,0 +1,539 @@
...
+defaultValue = "<%= parseEther('0.80') %>"
+
+[setting.low_utilization_slope_percent]
+defaultValue = "<%= parseEther('0.000025') %>"
+
+[setting.high_utilization_slope_percent]
+defaultValue = "<%= parseEther('0.01') %>"
+
+# ETH market configurations
+[setting.maker_fee]
```
noisekit, line 160:
```suggestion
[setting.maker_fee_eth]
```
just a nit but might be worth having things suffixed/prefixed?

noisekit: Considering this will be part of main deployment that may have other collaterals/tokens/perps?

omnibus-sepolia-carina.toml
```diff
@@ -0,0 +1,539 @@
...
+
+# --- Configure oracles for markets --- #
+#
+#
+
+# wstETH/stETH oracle
+[provision.wsteth_to_steth_ratio_oracle]
+source = "<%= settings.wsteth_to_steth_ratio_oracle %>"
+targetPreset = "<%= settings.target_preset %>"
+options.wsteth = "<%= settings.wsteth_address %>"
```
noisekit, line 230: maybe need owner and salt as well?
```suggestion
options.owner = "<%= settings.owner %>"
options.salt = "<%= settings.salt %>"
options.wsteth = "<%= settings.wsteth_address %>"
```

omnibus-sepolia-carina.toml
davidvuong: One market for now, changing to add suffix if we add more.

davidvuong: Isolated deployment file that we'll revisit and refactor once testnet working

omnibus-sepolia-carina.toml
```diff
@@ -0,0 +1,539 @@
...
+
+# --- Configure bfp --- #
+#
+#
+
+# --- Set pyth --- #
+#
+#
+#
+[invoke.set_pyth]
```
noisekit, line 297: As we will have this config part of global v3 config would be good to prefix invoke names accordingly.
Ideally these invoke names should not be needed at all, but as we still need them, my suggestion is to have names that are reflective of what they call (`ContractName_funcName`).

in this case would be 
```suggestion
[invoke.BfpMarketProxy_setPyth]
```
I have tried this approach in arbitrum deployments to avoid the need to figure out names.

omnibus-sepolia-carina.toml
```diff
@@ -0,0 +1,547 @@
...
+minKeeperFeeUsd = "<%= settings.min_keeper_fee_usd %>"
+maxKeeperFeeUsd = "<%= settings.max_keeper_fee_usd %>"
+keeperProfitMarginUsd = "<%= settings.keeper_profit_margin_usd %>"
+keeperProfitMarginPercent = "<%= settings.keeper_profit_margin_percent %>"
+keeperSettlementGasUnits = 1200000
+keeperCancellationGasUnits = 600000
+keeperFlagGasUnits = 1200000
+keeperLiquidateMarginGasUnits = 1200000
+keeperLiquidationGasUnits = 1200000
+keeperLiquidationEndorsed = "<%= settings.deployer %>"                          # TODO, this is currently owner, change to a keeper
```
noisekit, line 364:
```suggestion
keeperLiquidationEndorsed = "<%= settings.owner %>"                          # TODO, this is currently owner, change to a keeper
```

owner !== deployer so either comment is not correct or value.

omnibus-sepolia-carina.toml
```diff
@@ -0,0 +1,547 @@
...
+
+# --- Set eth oracle node id --- #
+#
+#
+
+[invoke.set_eth_oracle_node_id]
+target = ["bfp_market_factory.BfpMarketProxy"]
+fromCall.func = "owner"
+func = "setEthOracleNodeId"
+args = ["<%= settings.eth_oracle_id %>"]
```
noisekit, line 380: I cannot find eth_oracle_id setting anywhere.

davidvuong: Surprised this wasn't caught in the dry-run but good catch

0xjocke: We include `"tomls/oracles/eth.toml",` which defines it as an extra:
https://github.com/Synthetixio/synthetix-deployments/blob/bfp-market-carina/tomls/oracles/eth.toml#L14

omnibus-sepolia-carina.toml
```diff
@@ -0,0 +1,547 @@
...
+target = ["system.CoreProxy"]
+fromCall.func = "owner"
+func = "configureMaximumMarketCollateral"
+args = [
+    "<%= settings.eth_market_id %>",
+    "<%= settings.weth_address %>",
+    "<%= settings.weth_max_pool_collateral %>",
+]
+
+[invoke.wsteth_maximum_market_collateral]
```
noisekit, line 497: for core we would also need to add `configureMaximumMarketCollateralValue` that limits max market value in $

omnibus-sepolia-carina.toml
0xjocke: Updated comment

omnibus-sepolia-carina.toml
0xjocke: You want this prefex on all invokes that have target `bfp_market_factory.BfpMarketProxy`? Did you change `setPyth` to camelcase on purpose?

omnibus-sepolia-carina.toml
noisekit: I believe this is a bad idea to refer them as `settings` as they clearly are not settings, but values derived from events. I strongly suggest to refer them as `extras. eth_oracle_id`.

omnibus-sepolia-carina.toml
0xjocke: I dont think that is true. Core does not have any maximum defined in USD and `configureMaximumMarketCollateralValue` does not exists. 

I think you might be thinking about the `setMaxMarketValue` and `setMaxMarketSize` which are not core method but perpsv3 specific methods

noisekit (approved):
Consider my comments, but they are not blockers. Overall looks good to me.

omnibus-sepolia-carina.toml
0xjocke: > I believe this is a bad idea to refer them as `settings` as they clearly are not settings, but values derived from events. I strongly suggest to refer them as `extras. eth_oracle_id`.

my understanding is that extras is deprecated?

omnibus-sepolia-carina.toml
noisekit: I found this to be probably the easiest approach to name invokes. and add suffix for significant part if there are multiple invokes of same function.
this name should have been automatic, but as we have to write it manually - it is very annoying to "create" new names for every step, so the easiest approach I found is jut to blindly copypaste `Contract_funcName_maybeArgValue`. That also comes quite nicely in cannon UI and logs.

noisekit: image


see example here https://usecannon.com/packages/synthetix-omnibus/latest/421614-main

omnibus-sepolia-carina.toml
noisekit: > my understanding is that extras are deprecated?

I strongly disagree and will contest this decision

omnibus-sepolia-carina.toml
noisekit: You are right, I've mixed up perps v3 with this.

omnibus-sepolia-carina.toml
```diff
@@ -0,0 +1,517 @@
...
+options.synthetixPreset = "<%= settings.target_preset %>"
+depends = ["provision.system"]
+
+# --- Provision fake collaterals --- #
+#
+#
+# Create a mintable fake ETH for collateral
+[provision.weth_mock_collateral]
+source = "mintable-token:1.8"
+options.name = "Fake wETH"
```
davidvuong, line 224:
```suggestion
options.name = "Fake WETH"
```

omnibus-sepolia-carina.toml
```diff
@@ -0,0 +1,517 @@
...
+options.symbol = "fWETH"
+options.owner = "<%= settings.owner %>"
+options.decimals = "18"
+
+# Set weth_address so that Core is configured with the fake wETH.
```
davidvuong, line 229:
```suggestion
# Set weth_address so that Core is configured with the fake WETH.
```

davidvuong (approved):
sned it pls 🚀 🌔

PR #250, opened 2024-06-03
legacy market

dbeal-eth:
mainnet and sepolia legacy market deployment files


e2e/generateDeployments.js
```diff
@@ -75,6 +75,12 @@ async function run() {
   const system = deployments.state['provision.system'].artifacts.imports.system;
   const contracts = {};
 
+  contracts.LegacyMarketProxy =
+    deployments.state['provision.legacyMarket'].artifacts.imports.legacyMarket.contracts.Proxy;
```
noisekit, line 79: This does not exist on any other deployments. It needs to be conditioned. Also why is it added first and not last?

e2e/tests/LegacyMarket.js
```diff
@@ -0,0 +1,112 @@
+const assert = require('assert');
```
noisekit, line 1: Needs to be individual for each deployment. No common tests please

e2e/generateDeployments.js
```diff
@@ -74,12 +74,20 @@ async function run() {
...
-
   contracts.CoreProxy = system.contracts.CoreProxy;
   contracts.AccountProxy = system.contracts.AccountProxy;
   contracts.USDProxy = system.contracts.USDProxy;
   contracts.OracleManagerProxy = system.imports.oracle_manager.contracts.Proxy;
 
+  if (deployments.state['provision.legacyMarket']) {
+    contracts.LegacyMarketProxy =
+      deployments.state['provision.legacyMarket'].artifacts.imports.legacyMarket.contracts.Proxy;
+    contracts.V2xSnx =
```
noisekit, line 85: Would be good to avoid shortening of names and ideally follow contract name, so in this case it could be 
```suggestion
    contracts.V2xSynthetix =
```

e2e/tasks/setAllPoolsMinDelegationTimeZero.js
```diff
@@ -0,0 +1,45 @@
+#!/usr/bin/env node
+
+const { ethers } = require('ethers');
+const { parseError } = require('../parseError');
+const { gasLog } = require('../gasLog');
+
+const log = require('debug')(`e2e:${require('path').basename(__filename, '.js')}`);
+
+async function setAllPoolsMinDelegationTimeZero({ poolId }) {
```
noisekit, line 9: Making this smaller "task" will make it more reusable.

```suggestion
async function setMarketMinDelegationTime({ poolId, marketId, delegationTime }) {
```

This seems to be a specific stuff for legacy market migration so fetching markets and iterating over them and calling `setMarketMinDelegationTime` would suit the code of test itself better, than task.

This is more of a suggestion as tasks are meant to be more of a "reusable" smaller blocks of work that don't do more than one thing

e2e/tests/omnibus-sepolia.toml/LegacyMarket.e2e.js
```diff
@@ -0,0 +1,117 @@
...
+
+const log = require('debug')(`e2e:${require('path').basename(__filename, '.js')}`);
+
+const run = function () {
+  let LegacyMarketProxy;
+  let CoreProxy;
+  let V2xSnx;
+
+  // some big wallets
+  // https://etherscan.io/token/0x89FCb32F29e509cc42d0C8b6f058C993013A843F#balances
```
noisekit, line 15:
```suggestion
  // https://sepolia.etherscan.io/token/ ??? token address? ??? #balances
```

e2e/tests/omnibus-sepolia.toml/LegacyMarket.e2e.js
```diff
@@ -0,0 +1,117 @@
...
+  const accounts = [
+    '0x99F4176EE457afedFfCB1839c7aB7A030a5e4A92',
+    '0x8cA24021E3Ee3B5c241BBfcee0712554D7Dc38a1',
+    '0x27Cc4d6bc95b55a3a981BF1F1c7261CDa7bB0931',
+    '0xcb68110C43C97b6051FEd5e2Bacc2814aDaD1688',
+    '0x722f0a6F970F8B144F5e92aB3f3BeaFA280b282f',
```
noisekit, line 21: these are empty account on sepolia

omnibus-mainnet.toml
```diff
@@ -7,20 +7,24 @@ include = [
...
+    "tomls/markets/legacy.toml",
     # "tomls/oracles/eth.toml",
     # "tomls/collaterals/weth.toml",
     "tomls/pools/spartan-council.toml",
     "tomls/permit-all-transferCrossChain.toml",
 ]
 
 [setting.snx_package]
-defaultValue = "synthetix:3.3.5"
+defaultValue = "synthetix:3.3.15"
```
noisekit, line 18: Just confirming we have a SIP to upgrade mainnet.

omnibus-mainnet.toml
```diff
@@ -7,20 +7,24 @@ include = [
...
 
 [setting.spot_market_package]
-defaultValue = "synthetix-spot-market:3.3.5"
+defaultValue = "synthetix-spot-market:3.3.15"
 
 [setting.perps_market_package]
-defaultValue = "synthetix-perps-market:3.3.5"
+defaultValue = "synthetix-perps-market:3.3.15"
```
noisekit, line 24: 1. go with the latest?
```suggestion
defaultValue = "synthetix-perps-market:3.3.19"
```
2. we can actually remove this as we dont have perps deployed, or do we?

omnibus-mainnet.toml
noisekit, line 18: the latest is 18 now
```suggestion
defaultValue = "synthetix:3.3.18"
```

noisekit (commented):
Reminder to bump versions for sepolia and mainnet deployments.

e2e/tests/omnibus-sepolia.toml/LegacyMarket.e2e.js
dbeal-eth: yea I know. we need to put actual addresses for sepolia but i'm not sure of one

dbeal-eth: so focusing on testing mainnet

e2e/tasks/setAllPoolsMinDelegationTimeZero.js
dbeal-eth: this task would be difficult to "reuse" outside of testing context because in normal operation it can only be executed by the market contract itself.

and I cant think of a situation where we would not want to set all min delgation times to any value other than 0, especially since we are getting close to phasing this system out.

e2e/tests/omnibus-sepolia.toml/LegacyMarket.e2e.js
dbeal-eth: the addresses here are actually mainnet so I dont think its necessary to change the url here to something less useful?

omnibus-mainnet.toml
dbeal-eth: thats right we dont have any perps deployed

e2e/tasks/setAllPoolsMinDelegationTimeZero.js
noisekit: situation to use non-zero value is to set it back to non zero value after setting it to zero value =)

e2e/tests/omnibus-sepolia.toml/LegacyMarket.e2e.js
noisekit: it is a sepolia test so we need to either use legit addresses from sepolia, or remove this part of test entirely as it is functional

omnibus-mainnet.toml
```diff
@@ -50,7 +51,10 @@ defaultValue = "2000000000000000000"
 defaultValue = "86400"
 
 [setting.deployer]
-defaultValue = "0x302d2451d9f47620374B54c521423Bf0403916A2"
+defaultValue = "0xEde8a407913A874Dd7e3d5B731AFcA135D30375E"
```
noisekit, line 54: need to revert the change potentially?

noisekit (approved):

PR #266, opened 2024-06-17
SIP-392: Support Ethena USDe as LP Collateral

noisekit:


kaleb-keny:
probably best @noisekit to resolve the current issues , before proceeding with more deploys

noisekit:
> probably best @noisekit to resolve the current issues , before proceeding with more deploys

yeah def.

kaleb-keny:
Seems like the changes do a lot more than just adding ethena USDe?

noisekit:
> Seems like the changes do a lot more than just adding ethena USDe?

There are updates for tests. And also some minor updates of arb sepolia to bring it closer to the arb mainnet (remove perps market mainly)

tomls/omnibus-arbitrum-mainnet/collaterals/USDe.toml
```diff
@@ -0,0 +1,39 @@
...
+
+[invoke.CoreProxy_setPoolCollateralConfiguration_USDe]
+target = ["system.CoreProxy"]
+fromCall.func = "getPoolOwner"
+fromCall.args = ["<%= extras.spartan_council_pool_id %>"]
+func = "setPoolCollateralConfiguration"
+args = [
+    "<%= extras.spartan_council_pool_id %>",
+    "<%= settings.USDe_address %>",
+    { collateralLimitD18 = "<%= settings.USDe_max_collateral_limit %>", issuanceRatioD18 = 0 },
```
kaleb-keny, line 38: should issuanceRatioD18 be zero here?

kaleb-keny (approved):

tomls/omnibus-arbitrum-mainnet/collaterals/USDe.toml
kaleb-keny: am pretty sure it is, but not sure why 😅

noisekit: yes. When it is zero for collateral config, then default global value is used.

kaleb-keny:
@noisekit 

> > Seems like the changes do a lot more than just adding ethena USDe?
> 
> There are updates for tests. And also some minor updates of arb sepolia to bring it closer to the arb mainnet (remove perps market mainly)

https://github.com/Synthetixio/synthetix-deployments/tree/main/tomls/omnibus-arbitrum-mainnet-arbthetix

But this seems to affect mainnet? The reward toml files that are not USDe

noisekit:
> But this seems to affect mainnet? The reward toml files that are not USDe

Yes, I've just moved configs into separate files instead of keeping them in the main toml. No txns to execute.

omnibus-arbitrum-sepolia.toml
```diff
@@ -278,91 +308,3 @@ args = [
     18,
     "<%= imports.system.contracts.USDRouter.address %>",
 ]
-
-[provision.RewardsDistributor_SpartanCouncilPool_ARB_rewards_for_ARB_lp]
```
noisekit, line 282: @kaleb-keny better comment on code itself so we can have proper reply threads.

These configs were extracted into separate files as we were getting way to much of them at this point. This does not create any new transactions.

tomls/omnibus-arbitrum-mainnet/collaterals/USDe.toml
kaleb-keny: Should we have default value  or the sccp value?

noisekit: the "default" is a global value defined by `invoke.CoreProxy_configureCollateral_USDe` 
```
issuanceRatioD18 = "<%= settings.USDe_issuance_ratio %>"
```
on line 24 above. And it is as per SCCP - `200%`

This is a specific value for this collateral for pool 1, that does not need to override global value for this collateral.

kaleb-keny: ohhh ok, i thought you meant by global, as in 1 value that rules them all... this is more of a local value, if it's specific to the USDe collateral i think 🤔

PR #270, opened 2024-06-20
Move all new markets to the HighCap Supermarket (2/2)

leomassazza:
2nd part of the migration of all new markets to the old good supermarket. 

This one depends on the execution of https://github.com/Synthetixio/synthetix-deployments/pull/273


omnibus-base-sepolia-andromeda.toml
```diff
@@ -10,12 +10,10 @@ include = [
     "tomls/pools/spartan-council.toml",                             # Create Spartan Council Pool
     "tomls/markets/spot-factory.toml",
     "tomls/markets/perps-highcap-factory.toml",                     # First supermarket (High MCap)
-    "tomls/markets/perps-lowcap-factory.toml",                      # Second supermarket (Low MCap)
```
noisekit, line 13: Just a note here that removing it does not mean it will be "undeployed" so this perps market still exists in the system

omnibus-base-sepolia-andromeda.toml
```diff
@@ -10,12 +10,10 @@ include = [
...
     "tomls/markets/common/bigcap-settings.toml",
     "tomls/omnibus-base-sepolia-andromeda/spot/usdc.toml",          # sUSDC spot market/wrapper, mock for USDC
     "tomls/omnibus-base-sepolia-andromeda/spot/statausdc.toml",     # sStataUSDC spot market/wrapper, mock for aUSDC/stataUSDC from erc-4626-to-assets-ratio-oracle
     "tomls/permit-all-perps-perpsHighcapSystem.toml",               # Enable perps market
-    "tomls/permit-all-perps-perpsLowcapSystem.toml",                # Enable perps market
```
noisekit, line 18: Need rug access to this market for everyone, not just remove "allowAll"

kaleb-keny: it's fine because it's sepolia, right?

noisekit: fair enough, not a big deal yeah

omnibus-base-sepolia-andromeda.toml
leomassazza: thanks for noticing!

leomassazza: added it back with a comment that is DISABLED

omnibus-base-sepolia-andromeda.toml
leomassazza: already fixed in the previous PR

noisekit (approved):
I think we got some rounding issue in test, which is strange. I'd just restart it. Should pass I reckon.

PR #273, opened 2024-06-21
Remove SCPool support to 2nd supermarket and revoke permits (1/2)

leomassazza:
This is the first PR of two to move all new markets to the old good supermarket and not use a 2nd one.

In this step the 2nd supermarket is removed from SC Pool and the permits revoked, the next PR will change all the markets to use the main supermarket as proxy (add the markets to HighCap supermarket). 

Notice that the LowCap supermarket will still be around, but without any pool support or permissions so should be useless.


noisekit (approved):

kaleb-keny (approved):

omnibus-base-sepolia-andromeda.toml
```diff
@@ -1,5 +1,5 @@
 name = "synthetix-omnibus"
-version = "34"
+version = "35"
```
noisekit, line 2: I have deployed this now.

PR #274, opened 2024-06-24
Remove sUSDe

noisekit:


omnibus-arbitrum-mainnet.toml
```diff
@@ -26,11 +26,6 @@ include = [
     "tomls/omnibus-arbitrum-mainnet/collaterals/weth.toml",
     "tomls/omnibus-arbitrum-mainnet/rewards/ARB_rewards_for_WETH_lp.toml",
 
-    # sUSDe
-    "tomls/omnibus-arbitrum-mainnet/oracles/pyth-sUSDe.toml",
-    "tomls/omnibus-arbitrum-mainnet/collaterals/sUSDe.toml",
-    "tomls/omnibus-arbitrum-mainnet/rewards/ARB_rewards_for_sUSDe_lp.toml",
```
noisekit, line 32: We haven't deployed mainnet yet so we can still remove these sUSDe deployments. Still bumping version though.

omnibus-arbitrum-sepolia.toml
```diff
@@ -1,5 +1,5 @@
 name = "synthetix-omnibus"
-version = "10"
+version = "11"
```
noisekit, line 2: Not removing from sepolia because already deployed. Just flipping the switch to disable deposits.

noisekit:
because the pool changes in #273 we cannot build base sepolia for now without price updates =(

davidvuong (approved):

PR #276, opened 2024-06-25
Dependabot config for cannon

noisekit:


.github/dependabot.yml
```diff
@@ -0,0 +1,20 @@
+updates:
+  - package-ecosystem: "yarn"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    groups:
+      cannon:
```
noisekit, line 7: experimenting here. Dependabot will probably try to update all deps anyway. Need to adjust it as we go

PR #306, opened 2024-07-09
update staleness tolerance

kaleb-keny:


dbeal-eth (approved):
Worth noting this turned out to be the solution to the failing e2e tests

noisekit (commented):
Great. Thanks 🙏

PR #322, opened 2024-07-29
Set DAI leverage to 0

noisekit:


dbeal-eth (approved):
Looks good, but might make more sense to execute the dai deleverage outside of cannonfile since it's going to be thrown away afterwards effectively

PR #323, opened 2024-07-30
SCCP-344

kaleb-keny:


kaleb-keny:
Hey @dbeal-eth i've dropped the test you made on attempting migration on mainnet, because unfortunately all the accounts that could have been migrated by liquidation, have been migrated. Should i setup an issue in order to setup a state independent test? I.e. create random account, stake snx on v2x, move liq ratio around and liquidate? Or maybe this will need to be setup anyway, with the changes being done to the legacy account migration anyway, right?

dbeal-eth:
yea probably the best way to handle this would be to jack up the liquidation rate on v3 and migrate our test account

noisekit (approved):

PR #331, opened 2024-08-05
SCCP-2132: Reduce target staking ratio to 250%

dbeal-eth:


kaleb-keny:
Closing this as no longer relevant @dbeal-eth , let me know otherwise,

PR #333, opened 2024-08-06
New RewardsDistributor and LegacyMarket sepolia deploys

dbeal-eth:


omnibus-arbitrum-sepolia.toml
```diff
@@ -14,7 +14,6 @@ include = [
     # USDC
     "tomls/oracles/pyth-usdc.toml",
     "tomls/omnibus-arbitrum-sepolia/collaterals/usdc.toml",
-    "tomls/omnibus-arbitrum-sepolia/rewards/ARB_rewards_for_USDC_lp.toml",
```
noisekit, line 17: I think it is ok to keep these files with one invoke on them. We will have a good git diff showing how one rd replaced by another one instead of completely changing the structure. 
but im not too concerned with this

tomls/omnibus-arbitrum-sepolia/rewards/sc.toml
```diff
@@ -0,0 +1,65 @@
+[provision.RewardsDistributor_SpartanCouncilPool]
```
noisekit, line 1: This one needs all the new e2e tests added or existing ones updated

dbeal-eth: i will keep the old tests (since the old RDs are still in  play for a bit longer) and add new one as you suggest :+1:

kaleb-keny (approved):

PR #341, opened 2024-08-09
re-add old arb sepolia rewards files

dbeal-eth:
i dont agree with this personally since as of yet there has been no evidence that these files are necessary for anything downstream and there is no actual reason to think that, but might as well re-add them just in case I suppose.


noisekit (changes requested):
need to reference them in the main toml

noisekit (approved):

PR #347, opened 2024-08-13
Multi-collateral Perps liquidations  rewards on Arbitrum Sepolia

leomassazza:
Upgrade Arbitrum Sepolia to use the multi-collateral RD.

WIP Note: Needs to fix the spot and perps markets versions once the Multi-collateral v3 code is merged and published

PR #355, opened 2024-08-20
fix attempt

kaleb-keny:


tomls/omnibus-arbitrum-sepolia/perps/collaterals/btc.toml
```diff
@@ -41,7 +41,6 @@ args = [
         "<%= settings.dai_address %>",
         "<%= settings.usdc_address %>",
         "<%= settings.USDe_address %>",
-        "<%= settings.sUSDe_address %>",
```
noisekit, line 44: I'd expect `rewards_distributor_external_package` to change as well to use the latest version that introduced the change

kaleb-keny: https://github.com/Synthetixio/synthetix-deployments/pull/347/files#diff-90e83d999478043debde0c008cf77d10cdfa92c4e3be7da8c40755ab8f7a4bc1R91

kaleb-keny: could be a situation where having no collaterals deposited, leads to issues if hooked up to a distributor too... waiting for leo to be online to check it out.

PR #363, opened 2024-08-26
add multicall forwarder to Optimism-Mainnet omnibus

moss-eth:
This is a required pre-req for the gov v3 deployment. Important to get this merged asap !

### DIFF
--- a/omnibus-optimism-mainnet.toml
+++ b/omnibus-optimism-mainnet.toml
@@ -117,3 +117,8 @@ args = [
         { marketId = "<%= extras.synth_eth_market_id %>", weightD18 = 1, maxDebtShareValueD18 = "<%= parseEther('1') %>" },
     ],
 ]
+
+[clone.trusted_multicall_forwarder]
+source = "trusted-multicall-forwarder"
+target = "trusted-multicall-forwarder"
+options.salt = "<%= settings.salt %>"


barrasso (approved):

omnibus-optimism-mainnet.toml
```diff
@@ -117,3 +117,8 @@ args = [
         { marketId = "<%= extras.synth_eth_market_id %>", weightD18 = 1, maxDebtShareValueD18 = "<%= parseEther('1') %>" },
     ],
 ]
+
+[clone.trusted_multicall_forwarder]
+source = "trusted-multicall-forwarder"
```
noisekit, line 122: Please set the version and preset. Or we risk this being redeployed, which we don't want.

dbeal-eth (changes requested):
They should be already automatically deployed with the latest version of this system, so I don't understand why this PR is necessary.

noisekit:
> They should be already automatically deployed with the latest version of this system, so I don't understand why this PR is necessary.

We most likely need to bump the synthetix package version, as this is the old deployment and relies on old version of core

moss-eth:
> They should be already automatically deployed with the latest version of this system, so I don't understand why this PR is necessary.

Where is it deployed? I don't see it here https://usecannon.com/packages/synthetix-omnibus/latest/10-main

noisekit:
> Where is it deployed? I don't see it here https://usecannon.com/packages/synthetix-omnibus/latest/10-main

Because it is an old deployment with old core. Need to update synthetix package version.

moss-eth:
closing because it is not strictly necessary for us to run our test deployment

PR #367, opened 2024-08-28
Fix parseEthers usage on arb sepolia

noisekit:
no changes in the deployment as all values are exactly same
- remove `parseEthers` from invoke steps and only use them in settings
- cleanup `depends`


kaleb-keny (approved):

kaleb-keny:
appreciate the cleanup

PR #370, opened 2024-08-29
Gov v3 review

noisekit:


.circleci/config.yml
```diff
@@ -768,18 +768,18 @@ workflows:
           chain-id: 42161
           provider-url: https://arbitrum-mainnet.infura.io/v3/$INFURA_API_KEY
 
-      - fork-test:
+      - fork-test-new:
           name: test-snaxchain-testnet
           toml: omnibus-snaxchain-testnet.toml
-          package: "synthetix-omnibus:latest"
+          # package: "synthetix-omnibus:latest"
```
moss-eth, line 774: we should remove this commented out line and the one below right?

noisekit: no. after initial deployment - uncomment and change workflow from `fork-test-new` to `fork-test`

dbeal-eth (commented):
just a general note but seems like there are a lot of copied files that are exactly the same (and will likely want to remain exactly the same in the future).

considering there is al ot of cross chain going on here for gov, it seems unlikely we would ever want these settings (ex snapshot contract spec, registered emitters, gas limit) to diverge

PR #379, opened 2024-08-30
fix admin period

dbeal-eth:


tomls/omnibus-mainnet/governance/councils-satellite.toml
```diff
@@ -1,7 +1,7 @@
 [clone.governance_satellite_ambassador]
 source = "<%= settings.governance_satellite_package %>"
 targetPreset = "ambassador"
-options.salt = "<%= settings.salt %>-ambassador"
+options.salt = "<%= settings.salt %>-ambassador-2"
```
noisekit, line 4: won't it redeploy just because of `governance_administration_period_duration` change anyway? Seems unnecessary

dbeal-eth: i havent checked for sure but the `governance_administration_period_duration` is set on the `initOrUpgradeElectionModule` call, so it wont trigger a new proxy deployment

moss-eth: it can't reinitialize the proxy with this setting

moss-eth (approved):

tomls/omnibus-mainnet/governance/councils-satellite.toml
moss-eth: it can't reinitialize the proxy with this setting like db said

mjlescano: Yes, it is needed and is the idea, because we need to update the settings, but it should only trigger a re-deploy on the proxy and not the all the contracts.

PR #380, opened 2024-08-30
Update omnibus-optimism-mainnet.toml

barrasso:


dbeal-eth (changes requested):
the proxy addresses look the same to me, not sure why optimism shoudl be considered bricked...

dbeal-eth (approved):
ive been explained the situation.

noisekit (approved):

Issue #358, opened 2024-08-21
Check if what is registered is configured as margin


fredsnax:
the liquidation e2e doesn't check if what is registered is what is configured as margin, since we dont liquidate all margins

PR #393, opened 2024-09-04
Eligible margin liquidation

fredsnax:


fredsnax:
The issue added was to make sure an account has provided all the different margins in the system before calling liquidate to make sure that liquidations is not bricked. To test this I added a few things: 

First, I checked an account has no margin by calling getAccountCollateralIds({ accountId }) and then deposited to all 3 markets (code already existed) then checked that getAccountCollateralIds({ accountId }) has the same amount of entries returned as getSupportedCollaterals(); by checking the arrays are equal length ensuring the user has all active collateral types possible. this means if we change the config to add an extra collateral type for margin it will fail until we update the tests so the user has margin in every collateral possible

fredsnax:
ref: https://github.com/Synthetixio/synthetix-deployments/issues/358

e2e/tests/omnibus-arbitrum-sepolia.toml/Perps_trading.e2e.js
```diff
@@ -342,6 +345,13 @@ describe(require('path').basename(__filename, '.e2e.js'), function () {
       priceVerificationContract: extras.pyth_price_verification_address,
     });
 
+    const systemCollaterals = await getSupportedCollaterals();
+    const accountCollaterals = await getAccountCollateralIds({ accountId });
+    // NOTE this is a check that all margin in the system is provided by the account before liquidation tests
+    // to prevent the system from potentially being bricked
+    // ref: https://github.com/Synthetixio/synthetix-deployments/issues/358
+    assert.equal(systemCollaterals.length, accountCollaterals.length);
```
noisekit, line 353: Why do we check if user has all collaterals supported by the perps?

noisekit: Inotial* is hings are only needed when we want to compare new calues with old. For this test please add a separate "it", fetch, log and assert there in isolation

fredsnax: we need to make sure that liquidations are working across every margin collateral type. because Kaleb mentioned that if the configuration is not setup correctly then liquiidations might be bricked even if the deployment works fine. So in this case we are adding a check before the call to liquidate to make sure all the collaterals are in the user account before they are liquidated

issue: https://github.com/Synthetixio/synthetix-deployments/issues/358

I can put it in a separate it statement before liquidation but it is related to the liquidation. lmk what you think

noisekit: Yes but how do we make that account have them? Does it mean that any perps trader must deposit ALL perps collaterals before doing trading?
What is the mechanics of this from lets say frontend prospective?

noisekit: Full test implemented in https://github.com/Synthetixio/synthetix-deployments/pull/426

noisekit:
replaced by https://github.com/Synthetixio/synthetix-deployments/pull/426

PR #418, opened 2024-09-13
Bfp liquidation tests

0xMithrandir:
Add basic flagPosition/liquidate test for weth and susd.


e2e/tests/omnibus-sepolia.toml/BFP_Market_ETH.e2e.js
```diff
@@ -420,4 +420,130 @@ describe(require('path').basename(__filename, '.e2e.js'), function () {
...
+      contract: 'BfpMarketProxy',
+      func: 'flagPosition',
+      args: [accountId, marketId],
+    });
+
+    log({ tx, receipt });
+    log('Flagged position for liquidation');
+
+    // Attempt to liquidate. This should complete successfully.
+    await contractWrite({
```
noisekit, line 540: There is a limit on how long each "it" is executed for. And this is the reason all other tests have many "it" sections (one per action". Please split this case. And add some asserts too as now only logs are there.

e2e/tests/omnibus-sepolia.toml/BFP_Market_ETH.e2e.js
```diff
@@ -420,4 +420,130 @@ describe(require('path').basename(__filename, '.e2e.js'), function () {
...
+      wallet,
+      contract: 'BfpMarketProxy',
+      func: 'liquidatePosition',
+      args: [accountId, marketId],
+    });
+
+    log(accountId + ' position liquidated');
```
noisekit, line 547: log objects and not messages please

e2e/tests/omnibus-sepolia.toml/BFP_Market_ETH.e2e.js
0xMithrandir: I've split this in 2 parts and cleaned up unused bits

omnibus-sepolia.toml
```diff
@@ -452,3 +458,38 @@ args = [
     "<%= settings.weth_address %>",
     "<%= settings.weth_rewards_distributor %>",
 ]
+
+#  adding these for future reference, it should be ensured these are registered as well
```
noisekit, line 462: What about these calls? Why dont we enable them now?

e2e/tests/omnibus-sepolia.toml/BFP_Market_ETH.e2e.js
noisekit: Yeah thanks, looks ok

e2e/tests/omnibus-sepolia.toml/BFP_Market_ETH.e2e.js
```diff
@@ -420,4 +420,115 @@ describe(require('path').basename(__filename, '.e2e.js'), function () {
...
+
+      // start the liquidation process
+      const { tx, receipt } = await contractWrite({
+        wallet,
+        contract: 'BfpMarketProxy',
+        func: 'flagPosition',
+        args: [accountId, marketId],
+      });
+
+      log({ tx, receipt });
```
noisekit, line 521: tx info and receipt events are already logged in contractWrite. Not sure we need them here.

e2e/tests/omnibus-sepolia.toml/BFP_Market_ETH.e2e.js
noisekit: image

This is what I was referring to, takes over 20s.
Most likely due to waits. Maybe split a bit more and keep delays in a separate `it` case with telling caption?

omnibus-sepolia.toml
0xMithrandir: They've been manually executed on the sepolia testnet so cannon build breaks if I uncomment them here, I'll add these properly in the mainnet toml

e2e/tests/omnibus-sepolia.toml/BFP_Market_ETH.e2e.js
0xMithrandir: Removing them

omnibus-sepolia.toml
noisekit: Please talk to DB and use cannon `alter` to add these into the state without breaking. These are not the "transient" steps that should be executed without config changes. It is not a blocker for this PR though but please do follow up

e2e/tests/omnibus-sepolia.toml/BFP_Market_ETH.e2e.js
```diff
@@ -420,4 +420,113 @@ describe(require('path').basename(__filename, '.e2e.js'), function () {
     log({ newCollateral });
     assert.equal(newCollateral.toNumber(), 0);
   });
+
+  describe('Liquidation', async () => {
```
noisekit, line 424: Ok I have just realised this is a nested describe. Please extract liquidation tests into their own separate files and copy necessary "bootstrapping" `it` steps to it.

noisekit (changes requested):
Extract Liquidation tests into separate files entirely. Split to more `it` steps to avoid long execution of each `it` step

omnibus-sepolia.toml
0xMithrandir: Thanks, I didn't know about the cannon alter command.

e2e/tests/omnibus-sepolia.toml/BFP_Market_ETH.e2e.js
0xMithrandir: Made a new separate file for liquidation tests.

e2e/tests/omnibus-sepolia.toml/BFP_Market_Liquidations.e2e.js
```diff
@@ -0,0 +1,665 @@
...
+    assert.equal(await getAccountOwner({ accountId }), address);
+  });
+
+  it('should set fWETH balance to 1_000', async () => {
+    assert.equal(
+      await getCollateralBalance({ address, symbol: 'fWETH' }),
+      0,
+      'New wallet has 0 fWETH balance'
+    );
+    const { tokenAddress } = await getCollateralConfig('fWETH');
```
noisekit, line 88: unrelated to this PR at all
we do not need fWETH token. We can wrap normal eth. In all other tests in all other chains we do it with normal WETH contract and has special task to get yourself WETH (which calls WETHContract.deposit under the hood)

e2e/tests/omnibus-sepolia.toml/BFP_Market_Liquidations.e2e.js
```diff
@@ -0,0 +1,665 @@
...
+    });
+  });
+
+  it('should delegate 500 fWETH into the Spartan Council pool', async () => {
+    assert.deepEqual(await getAccountCollateral({ accountId, symbol: 'fWETH' }), {
+      totalDeposited: 500,
+      totalAssigned: 0,
+      totalLocked: 0,
+    });
+    await delegateCollateral({
```
noisekit, line 142: there is no need to deposit/delegate/etc to be able to trade on perps. 
Is there any particular reason we have LP functions here in the test for Perps liquidations?

My only guess is that pool is literally empty and there is no way to trade... Just guessing

e2e/tests/omnibus-sepolia.toml/BFP_Market_Liquidations.e2e.js
```diff
@@ -0,0 +1,665 @@
...
+      contract: 'BfpMarketProxy',
+      func: 'flagPosition',
+      args: [accountId, marketId],
+    });
+    let eventEmitted = false;
+    for (const event of events) {
+      if (event.event === 'PositionFlaggedLiquidation') {
+        eventEmitted = true;
+        console.assert(event.args.accountId.eq(accountId));
+        console.assert(event.args.marketId.eq(marketId));
```
noisekit, line 495: we dont have console.assert. Use `assert` instead

noisekit: Better to fill the array of necessary objects to check and do `assert.deepEqual` looking at full data here instead of list of small one-field asserts.
THis way in tests if there is an error we can see the nice diff between what's expected and what was actually there.

e2e/tests/omnibus-sepolia.toml/BFP_Market_Liquidations.e2e.js
```diff
@@ -0,0 +1,665 @@
...
+        {
+          marketId,
+          ...updatedMarketConfiguration,
+        },
+      ],
+      impersonate: owner,
+    });
+  });
+
+  it('should flag an underwater sUSD backed position', async () => {
```
noisekit, line 635: Do we need to have sUSD liquidation test together with ETH liquidation test? From what I can see it is a 1-1 copy for ETH. So unless there is a check here how these two separate positions affect each other - we should separate sUSD liquidation into its own file and reduce this one in half.

omnibus-sepolia.toml
noisekit: Given we now have the test to validate rewards distributors registrations - #426, current sepolia deployment will fail it and we definitely need to add these invokes to configs. Can be done in a separate PR (and the test from #426 must be added for BFP too - with its own flavour)

e2e/tests/omnibus-sepolia.toml/BFP_Market_Liquidations.e2e.js
noisekit: ok so I compared these two liquidations and they are absolutely identical. Please make a separate file for sUSD liquidation


Screenshot 2024-09-20 at 18 54 37

omnibus-sepolia.toml
0xMithrandir: I think it makes sense for this to go into a separate PR

e2e/tests/omnibus-sepolia.toml/BFP_Market_Liquidations.e2e.js
0xMithrandir: I've updated tests using `deepEqual`

noisekit (approved):
rebased and ready to go, thanks for the cleanup

noisekit:
arb fix is incoming #443, unrelated to this PR

noisekit:
image
oh cmon....

0xMithrandir:
> arb fix is incoming #443, unrelated to this PR

Only arbitrum-sepolia is left, when this arb fix is merged then this one can be merged as well.

PR #419, opened 2024-09-14
new oracle manager deployment

dbeal-eth:
enjoy.


tomls/omnibus-sepolia/core.toml
```diff
@@ -0,0 +1,26 @@
...
+[setting.target_preset]
+defaultValue = "main"
+
+[setting.salt]
+defaultValue = "snax"
+
+[setting.bundleSalt]
+defaultValue = "snax"
+
+[clone.oracle_manager]
```
noisekit, line 14: Do we need trusted multicall as well here somewhere?

noisekit (approved):

tomls/omnibus-sepolia/core.toml
dbeal-eth: It's not really necessary. If we were starting over then. Yeah it probably would be best to just have it on the top level, but it's not going to hurt anything and I don't foresee it hurting anything in the future because of the way that it's deployed. Basically functions like an import.

omnibus-base-sepolia-andromeda.toml
```diff
@@ -1,9 +1,9 @@
 name = "synthetix-omnibus"
-version = "49"
+version = "50"
```
noisekit, line 2: need to revert changes to base sepolia please

PR #431, opened 2024-09-20
Multicollateral-Perps-Base-Sepolia ⚾ 

kaleb-keny:


omnibus-base-sepolia-andromeda.toml
```diff
@@ -1,329 +1,202 @@
 name = "synthetix-omnibus"
-version = "49"
+version = "50"
```
noisekit, line 2: here is a readable diff for this new deploy: https://github.com/Synthetixio/Synthetix-Gitbook-v3/pull/450/files#diff-4a7da7a4a4b4551d5b607d27c46f12377f00b59ea969ad9187b388da157e3bd8

I did run over deploy steps and everything looks good so far (given the number of steps it is possible smth was missed).

So I ran over scripts to generate docs and contracts previews so we can have better readable diffs of what has changed.

There are plenty of values that need to be fixed to match existing deployment.

Make sure not to skip on large diff of `md` file as it is the most readable form at this moment.

kaleb-keny: resolved hopefully with the below prs

noisekit (approved):

PR #433, opened 2024-09-21
chore: bump cannon

saeta-eth:
Cannon v2.18.0 includes a new feature that validates that the settings used are defined, otherwise it will throw an error. These are the requested changes to make it work and attempt a smooth update. FYI: the values ​​used are fake, feel free to update them with the correct ones.


noisekit:
1. release cannon packages with `alpha` tag
2. use https://github.com/Synthetixio/synthetix-deployments/actions/workflows/cannon-update.yml workflow to trigger cannon `alpha` update

Let me know if anything does not work for you for some reason. If everything is working well, you may want to add this action to cannon docs.

noisekit:
The example of such automatically created PR: https://github.com/Synthetixio/synthetix-deployments/pull/442

PR #445, opened 2024-09-25
salt change

moss-eth:
due to error during deployment using mainnet deployer instead of testnet, some create2 addresses were taken and I had to restart the deployment

### DIFF
--- a/omnibus-snaxchain-testnet.toml
+++ b/omnibus-snaxchain-testnet.toml
@@ -9,7 +9,7 @@ include = [
 ]
 
 [setting.salt]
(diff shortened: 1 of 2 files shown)


barrasso (approved):

omnibus-snaxchain-testnet.toml
```diff
@@ -9,7 +9,7 @@ include = [
 ]
 
 [setting.salt]
-defaultValue = "better salt"
+defaultValue = "even better salt"
```
noisekit, line 12: kek

PR #455, opened 2024-10-02
Multicollateral-Perps-Base-Mainnet-Part-1

kaleb-keny:


noisekit (approved):
Looks good.
We need to downgrade oracle manager to pre-batch errors version which is outside of this PR change and needs to be done separately

PR #463, opened 2024-10-04
fix-strict-stalenss-multicollateral-deploy-base-sepolia

kaleb-keny:


tomls/omnibus-base-sepolia-andromeda/perps/markets/aave.toml
```diff
@@ -30,7 +30,7 @@ func = "updatePriceData"
 args = [
     "<%= settings.aave_perps_market_id %>",
     "<%= extras.aave_oracle_id %>",
-    "<%= settings.big_cap_strict_staleness_tolerance %>",
+    "<%= settings.bigCapDefaultStalenessTolerance %>",
```
noisekit, line 33: My suggestion is to set `big_cap_strict_staleness_tolerance` instead in `omnibus-base-sepolia-andromeda.toml`

kaleb-keny: yeah was about to make these changes, however, push an update to the common file will lead to conflict with the mainnet PR's so probably we tackle these after the mainnet PR's are merged and sent?

kaleb-keny: So 1 large PR change that includes these, post next week's release

noisekit (approved):

PR #481, opened 2024-10-11
bump tolerances

dbeal-eth:
this ensures that there is a one week tolerance for base sepolia nad arbitrum

NOTE: one of the perps packages appears to be downgraded. This is because it undoes the batched errors ERC7412 changes. for now this is fine, it was only on testnet and we will bring it back to testnet next week. This is mostly to accelerate the process of testing, verifying, and pushing the one month tolerance of prices.

### DIFF
--- a/omnibus-arbitrum-sepolia.toml
+++ b/omnibus-arbitrum-sepolia.toml
@@ -154,7 +154,7 @@ defaultValue = "arbitrum-gas-price-oracle:3.3.16"
 defaultValue = "synthetix-spot-market:3.7.1"
 
 [setting.perps_market_package]
-defaultValue = "synthetix-perps-market:3.7.1"
+defaultValue = "synthetix-perps-market:3.6.5"
 
 [setting.buyback_snx_package]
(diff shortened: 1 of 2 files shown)


kaleb-keny (approved):

omnibus-arbitrum-sepolia.toml
```diff
@@ -154,7 +154,7 @@ defaultValue = "arbitrum-gas-price-oracle:3.3.16"
 defaultValue = "synthetix-spot-market:3.7.1"
 
 [setting.perps_market_package]
-defaultValue = "synthetix-perps-market:3.7.1"
+defaultValue = "synthetix-perps-market:3.6.5"
```
noisekit, line 157: Why downgrade ?

noisekit: We used higher version because it had batch errors update

noisekit: Ok never mind read description

PR #504, opened 2024-10-21
SCCP-357-Base-Sepolia

kaleb-keny:


omnibus-base-sepolia-andromeda.toml
```diff
@@ -301,6 +326,44 @@ defaultValue = "0x48914229deDd5A9922f44441ffCCfC2Cb7856Ee9"
 defaultValue = "<%= imports.erc_4626_to_assets_ratio_oracle.imports.statausdc_token_mock.contracts.Token.address %>"
 description = "We use the mock generated by erc-4626-to-assets-ratio-oracle package in statausdc-usdc.toml"
 
+[provision.cbbtc_mock_collateral]
+source = "mintable-token:1.8"
+target = "synthetix-mock-tokens:1.8@cbbtc"
+options.name = "Fake Coinbase Bitcoin"
+options.symbol = "fcbbtc"
```
noisekit, line 333: to avoid diffs in tests, please use the same symbol for fake tokens as for real ones. So gonna be 
```suggestion
options.symbol = "cbBTC"
```
(same for other mocks)

omnibus-base-sepolia-andromeda.toml
```diff
@@ -301,6 +326,44 @@ defaultValue = "0x48914229deDd5A9922f44441ffCCfC2Cb7856Ee9"
...
+
+[provision.wsteth_mock_collateral]
+source = "mintable-token:1.8"
+target = "synthetix-mock-tokens:1.8@wsteth"
+options.name = "Fake Lido Wrapped Staked Eth"
+options.symbol = "fwsteth"
+options.owner = "<%= settings.owner %>"
+options.decimals = "18"
+
+[provision.weth_mock_collateral]
```
noisekit, line 353: we dont need WETH because we can get unlimited ETH and it is one `.deposit()` away from WETH. So please use real WETH

omnibus-base-sepolia-andromeda.toml
kaleb-keny: hey @noisekit why are we dropping the `f`  for fake?

omnibus-base-sepolia-andromeda.toml
kaleb-keny: would you mind if we keep it as fake? because i hate using real scarce tokens on testnets 😅 , but no preference if you have a strong opinion about it.

tomls/omnibus-base-sepolia-andromeda/rewards/scbbtc_perps_liquidations_rewards.toml
```diff
@@ -0,0 +1,46 @@
+[provision.RewardsDistributor_scbbtc_liquidation_rewards]
+source = "synthetix-rewards-dist-ext:3.6.0"
+target = "synthetix-rewards-dist-ext:3.6.0@perps-scbbtc"
+options.salt = "<%= settings.salt %>"
```
noisekit, line 4:
```suggestion
options.salt = "<%= settings.salt %>-perps-scbbtc"
```
Also need to change salt =|

tomls/omnibus-base-sepolia-andromeda/spot/cbeth.toml
```diff
@@ -0,0 +1,92 @@
+[var.cbeth_spot_settings]
+synth_cbeth_max_collateral_amount = "<%= parseEther(String(1_750)) %>"
+synth_cbeth_skew_scale = "<%= parseEther(String(350_000)) %>"
+
+[invoke.SpotMarketProxy_createSynth_cbeth]
+target = ["spotFactory.SpotMarketProxy"]
+fromCall.func = "owner"
+func = "createSynth"
+args = ["Synthetic Coinbase Wrapped Staked ETH", "sCBETH", "<%= settings.owner %>"]
```
noisekit, line 9:
```suggestion
args = ["Synthetic Coinbase Wrapped Staked ETH", "scbETH", "<%= settings.owner %>"]
```

tomls/omnibus-base-sepolia-andromeda/spot/cbbtc.toml
```diff
@@ -0,0 +1,92 @@
+[var.cbbtc_spot_settings]
+synth_cbbtc_max_collateral_amount = "<%= parseEther(String(100)) %>"
+synth_cbbtc_skew_scale = "<%= parseEther(String(35_000)) %>"
+
+[invoke.SpotMarketProxy_createSynth_cbbtc]
+target = ["spotFactory.SpotMarketProxy"]
+fromCall.func = "owner"
+func = "createSynth"
+args = ["Synthetic Coinbase Bitcoin", "sCBBTC", "<%= settings.owner %>"]
```
noisekit, line 9:
```suggestion
args = ["Synthetic Coinbase Wrapped BTC", "scbBTC", "<%= settings.owner %>"]
```
from https://help.coinbase.com/en/coinbase/trading-and-funding/sending-or-receiving-cryptocurrency/coinbase-wrapped-btc

tomls/omnibus-base-sepolia-andromeda/spot/wsteth.toml
```diff
@@ -0,0 +1,92 @@
+[var.wsteth_spot_settings]
+synth_wsteth_max_collateral_amount = "<%= parseEther(String(1_750)) %>"
+synth_wsteth_skew_scale = "<%= parseEther(String(350_000)) %>"
+
+[invoke.SpotMarketProxy_createSynth_wsteth]
+target = ["spotFactory.SpotMarketProxy"]
+fromCall.func = "owner"
+func = "createSynth"
+args = ["Synthetic Lido Wrapped Staked ETH", "sWSTETH", "<%= settings.owner %>"]
```
noisekit, line 9:
```suggestion
args = ["Synthetic Lido Wrapped Staked ETH", "wstETH", "<%= settings.owner %>"]
```

also technically token name is `Wrapped liquid staked Ether 2.0` from contract https://etherscan.io/token/0x7f39c581f595b53c5cb19bd0b3f8da6c935e2ca0#readContract

tomls/omnibus-base-sepolia-andromeda/spot/cbbtc.toml
kaleb-keny: yeah this is often confusing the casing needs to be right, i agree, but with an `s` there makes these decisions a bit more complicated tbh, so i would just like to uniformize these, in a way, where the actual asset in upper case and the protocol/cex it comes from in lower... what do you think? but will add that wrapped... Ofc if you think otherwise strongly happy to change, i will not die on this hill 😅

tomls/omnibus-base-sepolia-andromeda/spot/wsteth.toml
kaleb-keny: No worries sir will add the 2.0 😅  although what is 1.0

omnibus-base-sepolia-andromeda.toml
kaleb-keny: updated as well as others, but kept `f` can remove if you think worthwhile

omnibus-base-sepolia-andromeda.toml
noisekit: weth should not be fake. All testnets use real weth.

omnibus-base-sepolia-andromeda.toml
noisekit: Please remove `f`. That is pain to maintain separate symbols. Ideally e2e test should be pretty much exactly same as prod. And all configs as well. Only addresses different

kaleb-keny: ok will remove them nikita

omnibus-base-sepolia-andromeda.toml
kaleb-keny: i want to maintain this, i feel the pain of acquiring testnet tokens...

tomls/omnibus-base-sepolia-andromeda/spot/wsteth.toml
kaleb-keny: Well that almost caused an issue @noisekit , thankfully fixed 😅  missing `s` sir

omnibus-base-sepolia-andromeda.toml
noisekit: It probably is down to our context. I do all the testing in anvil. So getting weth is just `.deposit({value: whatever})` (because we have unlimited ETH) and all our tests are using real ETH -> WETH.

At the same time it is one of the "real" tokens that we can make tests as close to mainnet as possible (no magic).

So at least we can guaranteed have one real token in our testnet deployment,

These are my thoughts pro real WETH.

The only negative in using real WETH I could see if we do extensive testing on real sepolia and not forks, but I don't do it anywhere.

kaleb-keny: yeah alright please let's keep this as fake weth 🙏🏼  it helps those who need to use the chain for testing, has no net negative, the supply of these tokens is being constrained.... so hopefully no one is abusing anymore

kaleb-keny: But if you really really feel otherwise about it, happy to make the necessary updates ofc,

noisekit: I realised there are  more things. 1. Frontend cannot rely on weth being weth and do normal wrapping workflow and will have to treat it differently for testnet. 
2. e2e tests will need to differ for testnet and mainnet for weth in workflow again - testnet will have to use fake token mint instead of deposit. And so far weth tests were almost the only ones that are same for both testnet and mainnet

kaleb-keny: alright will revise then, no worries,

noisekit (approved):
As long as tests pass looks good

PR #513, opened 2024-10-25
Fix tbtc / btc price feed

kaleb-keny:
this is a temporary fix, until a long term fix where real is swapped to just btc and the spot things are tbtc
what it should do:
- use btc pyth feed in the perp market  for settlement of trades
- keep using tbtc pyth feed in the spot market for settlement of trades
- keep using tbtc pyth feed in the perp market to compute value of margin


noisekit (approved):
Would be better call old one tbtc and this one btc
Frontend uses this ne to infer tovken name so it will come up as "real_btc" symbol. 
it is not a big issue as this is not used as lp collateral but can potentially has some impact.

But we can rename settings later in the next pr if urgent.

PR #519, opened 2024-10-28
Change default staleness for testing

kaleb-keny:


kaleb-keny:
will revert these changes end of day

PR #587, opened 2024-12-11
Fix extraction of TrustedMulticallForwarder

noisekit:


dbeal-eth (approved):
Seems good but can probably only rely on the Oracle manager tmf because I don't expect that will be going anywhere

PR #682, opened 2025-04-11
Update Version - Mainnet.tom

kaleb-keny:
Given that version 18 was published, reverting back that version on the toml file.

### DIFF
--- a/omnibus-mainnet.toml
+++ b/omnibus-mainnet.toml
@@ -1,5 +1,5 @@
 name = "synthetix-omnibus"
-version = "19"
+version = "18"
 description = "Includes the full synthetix system with configurations applied"
 deployers = [
     "0x1C8236B406911A376369e33D39189F1b4B39F27D",


omnibus-mainnet.toml
```diff
@@ -1,5 +1,5 @@
 name = "synthetix-omnibus"
-version = "19"
+version = "18"
```
noisekit, line 2: But it is here https://github.com/Synthetixio/synthetix-deployments/commit/e8fe329a1669ea5853c236a7397da1d1a1b2aef0

was it made directly?

noisekit (approved):

omnibus-mainnet.toml
noisekit: Nevermind. I see the change here https://github.com/Synthetixio/Synthetix-Gitbook-v3/pull/671#pullrequestreview-2762124261

PR #694, opened 2025-05-21
update eos price feed to snx feed

kaleb-keny:
hey @noisekit i've updated the price feed of eos here to the snx  price feed. This is fine, because there are no positions and the caps are set to zero. https://exchange.synthetix.io/market/?asset=EOS&provider=snx_v3_base Once we have an `A` feed we'll switch it, but we need to send this transaction before the 25th in order to avoid having stale feeds... Since pyth is deprecating their EOS feed.


noisekit (approved):

kaleb-keny:
Correction, pyth is deprecating on the 23rd not the 25th

PR #709, opened 2025-06-12
Remove extra CollateralMock import

0xMithrandir:


e2e/tests/omnibus-sepolia.toml/test/Common.sol
```diff
@@ -5,13 +5,13 @@ import {CannonDeploy} from "./CannonDeploy.sol";
 import {ICoreProxy} from "deployments/sol/ICoreProxy.sol";
 import {IAccountProxy} from "deployments/sol/IAccountProxy.sol";
 import { IV4PerpsMarketProxy } from "deployments/sol/IV4PerpsMarketProxy.sol";
-import { CollateralMock } from "@synthetixio/main/contracts/mocks/CollateralMock.sol";
+import {IUSDProxy} from "deployments/sol/IUSDProxy.sol";
```
noisekit, line 8: IUSDProxy is snxUSD contract so it would not be correct to use it as collateral.

since forge does not have mocks anymore, it is ok to use `synthetixio/main/contracts/mocks/CollateralMock.sol` for tests only

PR #710, opened 2025-06-13
Disable deposits on Base

noisekit:


kaleb-keny (approved):
thank you for the memories

PR #713, opened 2025-06-29
lower oi caps to zero - base

kaleb-keny:


noisekit (approved):
Need to update omnibus version.

noisekit:
Our tests for base are disabled now, fyi

kaleb-keny:
> Need to update omnibus version.

awesome thanks for the quick check, done, alsow switched usdc and stata synth to enabled, since we didn't push that release the normal way

